Issue #1, opened 2019-05-27
Expand interface to allow route groups


samdark:

samdark:
1. Should support grouping multiple rules.
2. Group could be prefixed.
3. Group could have extra before/after middlewares.

Point 3, likely, isn't needed if we'll introduce a middleware for that in the yiisoft/web.

PR #3, opened 2019-05-28
Added groups

samdark:
| Q             | A
| ------------- | ---
| Is bugfix?    | no
| New feature?  | yes
| Breaks BC?    | no
| Tests pass?   | yes
| Fixed issues  | #1

### DIFF
--- a/src/Group.php
+++ b/src/Group.php
@@ -1,4 +1,5 @@
 <?php
+
 namespace Yiisoft\Router;
 
 class Group
--- a/src/Group.php
+++ b/src/Group.php
@@ -46,7 +46,7 @@ class Group
         return $this->prefix;
     }
 
-    public function getMiddleware(): GroupMiddlewareInterface
+    public function getMiddleware(): ?GroupMiddlewareInterface
     {
         return $this->middleware;
     }
--- a/src/Group.php
+++ b/src/Group.php
@@ -26,8 +26,11 @@ class Group
      */
     public function getRoutes(): iterable
     {
-        foreach ($this->routes as $route) {
-            yield $route->pattern($this->prefix . $route->getPattern());
-        }
+        return $this->routes;
+    }
+
+    public function getPrefix(): string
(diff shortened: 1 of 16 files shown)


src/Group.php
```diff
@@ -0,0 +1,30 @@
...
+    }
+
+    public function addRoute(Route $route): self
+    {
+        // TODO: immutable?
+        $this->routes[] = $route;
+        return $this;
+    }
+
+    public function getRoutes(): iterable
```
xepozz: It needs a PHPDoc with returning type `Route[]`

src/Group.php
```diff
@@ -0,0 +1,30 @@
+<?php
+namespace Yiisoft\Router;
+
+class Group
```
xepozz: How about test for Group?

samdark: Well, I think this one is unlikely to be merged. It's better to think on how to support native groups that are implemented in many routers.

xepozz: It can be implemented in adapter.

```php
class FastRouteAdapter implements \Yiisoft\Router\RouterInterface
{
    public function addGroup(\Yiisoft\Router\Group $group) 
    {
        $this->fastRoute->addGroup($group->getPrefix(), function (RouteCollector $r) use ($group) {
            foreach ($group->getRoutes() as $route) {
                $r->addRoute($route->getMethod(), $route->getPattern(), $route->getMiddleware());
            }
        });
    }
}
```

samdark: Re-checked. You're right.

samdark: There was a test for group class.

rustamwin:
No support middleware for the group?

samdark:
https://github.com/yiisoft/router-fastroute/pull/6

zabachok:
@samdark why you don't use typed properties?

samdark:
Was no time to convert to these in this case. Do you want to help?

PR #17, opened 2019-12-20
Add property types and make minor fixes

viktorprogger:
Add typed properties and fix some inspections of PhpStorm and its plugin [EA Extended](https://plugins.jetbrains.com/plugin/7622-php-inspections-ea-extended-/).

| Q             | A
| ------------- | ---
| Is bugfix?    | ❌
| New feature?  | ❌
| Breaks BC?    | ❌
| Tests pass?   | ✔️
| Fixed issues  | -

### DIFF
--- a/src/Group.php
+++ b/src/Group.php
@@ -2,14 +2,17 @@
 
 namespace Yiisoft\Router;
 
+use InvalidArgumentException;
 use Psr\Http\Server\MiddlewareInterface;
 use Yiisoft\Router\Middleware\Callback;
 
+use function is_callable;
+
 class Group implements RouteCollectorInterface
 {
-    protected $items = [];
-    protected $prefix;
-    protected $middlewares = [];
+    protected array $items = [];
+    protected string $prefix;
+    protected array $middlewares = [];
 
     public function __construct(string $prefix = null, callable $callback = null)
     {
@@ -32,12 +35,12 @@ class Group implements RouteCollectorInterface
 
     final public function addMiddleware($middleware): self
     {
-        if (\is_callable($middleware)) {
+        if (is_callable($middleware)) {
             $middleware = new Callback($middleware);
         }
 
         if (!$middleware instanceof MiddlewareInterface) {
-            throw new \InvalidArgumentException('Parameter should be either a PSR middleware or a callable.');
+            throw new InvalidArgumentException('Parameter should be either a PSR middleware or a callable.');
         }
 
         $this->middlewares[] = $middleware;
--- a/src/Group.php
+++ b/src/Group.php
@@ -5,7 +5,6 @@ namespace Yiisoft\Router;
 use InvalidArgumentException;
 use Psr\Http\Server\MiddlewareInterface;
 use Yiisoft\Router\Middleware\Callback;
-
 use function is_callable;
 
 class Group implements RouteCollectorInterface
@@ -14,7 +13,7 @@ class Group implements RouteCollectorInterface
     protected ?string $prefix;
     protected array $middlewares = [];
 
-    public function __construct(string $prefix = null, callable $callback = null)
+    public function __construct(?string $prefix = null, ?callable $callback = null)
     {
         $this->prefix = $prefix;
 
--- a/src/Group.php
+++ b/src/Group.php
@@ -5,7 +5,6 @@ namespace Yiisoft\Router;
 use InvalidArgumentException;
 use Psr\Http\Server\MiddlewareInterface;
 use Yiisoft\Router\Middleware\Callback;
-use function is_callable;
 
 class Group implements RouteCollectorInterface
 {
--- a/src/Group.php
+++ b/src/Group.php
@@ -32,6 +32,10 @@ class Group implements RouteCollectorInterface
         $this->items[] = new Group($prefix, $callback);
     }
 
+    /**
+     * @param callable|MiddlewareInterface $middleware
+     * @return $this
+     */
     final public function addMiddleware($middleware): self
     {
         if (is_callable($middleware)) {
--- a/src/Group.php
+++ b/src/Group.php
@@ -11,7 +11,7 @@ use function is_callable;
 class Group implements RouteCollectorInterface
 {
     protected array $items = [];
-    protected string $prefix;
+    protected ?string $prefix;
     protected array $middlewares = [];
 
     public function __construct(string $prefix = null, callable $callback = null)
--- a/src/MatchingResult.php
+++ b/src/MatchingResult.php
@@ -57,7 +57,7 @@ final class MatchingResult implements MiddlewareInterface
 
     public function process(ServerRequestInterface $request, RequestHandlerInterface $handler): ResponseInterface
     {
-        if ($this->success === false || $this->route === null) {
+        if ($this->success === false) {
             return $handler->handle($request);
         }
 
--- a/src/MatchingResult.php
+++ b/src/MatchingResult.php
@@ -9,11 +9,7 @@ use Psr\Http\Server\RequestHandlerInterface;
 
 final class MatchingResult implements MiddlewareInterface
 {
-    private ?bool $success;
-
-    /**
-     * @var Route
-     */
+    private bool $success;
     private Route $route;
     private array $parameters = [];
     private array $methods;
@@ -24,7 +20,7 @@ final class MatchingResult implements MiddlewareInterface
 
     public static function fromSuccess(Route $route, array $parameters): self
     {
-        $new = new self();
+        $new = new static();
         $new->success = true;
         $new->route = $route;
         $new->parameters = $parameters;
--- a/src/MatchingResult.php
+++ b/src/MatchingResult.php
@@ -20,7 +20,7 @@ final class MatchingResult implements MiddlewareInterface
 
     public static function fromSuccess(Route $route, array $parameters): self
     {
-        $new = new static();
+        $new = new self();
         $new->success = true;
         $new->route = $route;
         $new->parameters = $parameters;
@@ -29,7 +29,7 @@ final class MatchingResult implements MiddlewareInterface
 
     public static function fromFailure(array $methods): self
     {
-        $new = new static();
+        $new = new self();
         $new->methods = $methods;
         $new->success = false;
         return $new;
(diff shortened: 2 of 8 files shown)


src/MatchingResult.php
```diff
@@ -9,14 +9,14 @@
 
 final class MatchingResult implements MiddlewareInterface
 {
-    private $success;
+    private ?bool $success;
 
     /**
      * @var Route
```
xepozz: It's useless now

src/RouteNotFoundException.php
```diff
@@ -2,11 +2,12 @@
...
 namespace Yiisoft\Router;
 
+use RuntimeException;
 use Throwable;
 
-final class RouteNotFoundException extends \RuntimeException
+final class RouteNotFoundException extends RuntimeException
 {
-    public function __construct($routeName = '', $code = 0, Throwable $previous = null)
+    public function __construct(?string $routeName = '', ?int $code = 0, Throwable $previous = null)
```
xepozz: Why `$routeName` and `$code` is nullable?

src/RouterFactory.php
```diff
@@ -2,14 +2,26 @@
...
 
+use function count;
+use function is_array;
+use function is_callable;
+use function is_string;
+
 final class RouterFactory
 {
+    /**
+     * @var callable
```
xepozz: How about typing the property?

src/MatchingResult.php
```diff
@@ -9,14 +9,14 @@
 
 final class MatchingResult implements MiddlewareInterface
 {
-    private $success;
+    private ?bool $success;
```
xepozz: Why it's nullable?

src/Group.php
```diff
@@ -2,14 +2,17 @@
...
 
+use function is_callable;
+
 class Group implements RouteCollectorInterface
 {
-    protected $items = [];
-    protected $prefix;
-    protected $middlewares = [];
+    protected array $items = [];
+    protected string $prefix;
```
samdark: Prefix is nullable (see constructor).

src/MatchingResult.php
```diff
@@ -9,14 +9,14 @@
...
 
     /**
      * @var Route
      */
-    private $route;
-    private $parameters = [];
-    private $methods;
+    private Route $route;
+    private array $parameters = [];
+    private array $methods;
```
samdark: Is there a reason not to initialize it?

src/MatchingResult.php
```diff
@@ -9,14 +9,14 @@
...
-    private $success;
+    private ?bool $success;
 
     /**
      * @var Route
      */
-    private $route;
-    private $parameters = [];
-    private $methods;
+    private Route $route;
```
samdark: Is there a reason not to initialize it?

samdark:
https://travis-ci.com/yiisoft/router/builds/141881228

src/RouterFactory.php
viktorprogger: Properties don't support `callable` type for now

src/RouteNotFoundException.php
viktorprogger: There is really no reason for it

src/MatchingResult.php
viktorprogger: Is it a good idea to initialize it with `Method::ANY`?

src/RouterFactory.php
xepozz: What about `\Closure`?

src/MatchingResult.php
viktorprogger: 1. For some reason (I don't know why) it is not initialized when creating in the `fromFailure` method. But there will be an error if smbdy will try to `process` this object without `route` property.
1. This property is filled with some some object when creating a `MatcingResult` with `fromSuccess`. And we will spend some extra time and memory if we will fill it with wome another value before this.

Conclusion: this property must be filled, but I don't know the best place and the best way to od this.

src/MatchingResult.php
samdark: Yes, that's fine.

viktorprogger:
Any suggestions what does `styleci` want? Doesn't it understand field typehints?

src/MatchingResult.php
```diff
@@ -9,22 +9,18 @@
...
+    private array $methods = Method::ANY;
 
     private function __construct()
     {
     }
 
     public static function fromSuccess(Route $route, array $parameters): self
     {
-        $new = new self();
+        $new = new static();
```
viktorprogger: Please have an attension here. Is this ok ot there was some reason to leave `self`?

src/MatchingResult.php
samdark: Re-checked. This one is OK as is. It's used in `process()` and there's a check necessary to prevent error.

src/MatchingResult.php
samdark: The class is final so `static` doesn't make sense.

samdark:
StyleCI doesn't work with PHP 7.4 yet.

src/MatchingResult.php
viktorprogger: Made `self` in all cases here

src/MatchingResult.php
viktorprogger: Made a check there. Look at it please.

src/Route.php
```diff
@@ -2,56 +2,36 @@
...
 /**
  * Route defines a mapping from URL to callback / name and vice versa
  */
 final class Route implements MiddlewareInterface, RequestHandlerInterface
 {
+    private string $name;
+    /** @var string[] */
+    private array $methods;
+    private string $pattern;
+    private string $host;
```
xepozz: Host is nullable :) [See](https://github.com/yiisoft/router/blob/a71a39c7faf22cf2db8dfba01e8bfa23198b9ec7/src/Route.php#L262)

src/Route.php
```diff
@@ -2,56 +2,36 @@
...
-     * @var array
-     */
-    private $defaults = [];
-
-    /**
-     * @var RequestHandlerInterface
+     * @var MiddlewareInterface[]|callable[]
      */
-    private $nextHandler;
+    private array $middlewares;
```
xepozz:
```suggestion
    private array $middlewares = [];
```

src/Route.php
viktorprogger: [Can't agree with you](https://github.com/yiisoft/router/blob/a71a39c7faf22cf2db8dfba01e8bfa23198b9ec7/src/Route.php#L126). It can be null only while it is not set at all.

rob006: @viktorprogger Which makes it nullable. You know that accessing  initialized property will produce an error? If there is a place where `null` is acceptable value of property, then this property should be nullable and initialized as `null` by default.

xepozz: @viktorprogger Do you see call method `host` [there](https://github.com/yiisoft/yii-demo/blob/master/src/Factory/AppRouterFactory.php)?

viktorprogger: Thnx, @rob006, didn't know that. Will be more carefull.

src/MatchingResult.php
```diff
@@ -61,7 +57,7 @@ public function methods(): array
 
     public function process(ServerRequestInterface $request, RequestHandlerInterface $handler): ResponseInterface
     {
-        if ($this->success === false) {
+        if ($this->success === false || $this->route === null) {
```
samdark: There's no need for this check. It will be non-null if success is `true`.

src/Group.php
```diff
@@ -2,16 +2,18 @@
...
 {
-    protected $items = [];
-    protected $prefix;
-    protected $middlewares = [];
+    protected array $items = [];
+    protected ?string $prefix;
+    protected array $middlewares = [];
 
-    public function __construct(string $prefix = null, callable $callback = null)
+    public function __construct(?string $prefix = null, ?callable $callback = null)
```
samdark: There's no need making these nullable.

src/Route.php
```diff
@@ -2,56 +2,36 @@
...
+use function is_callable;
+use function next;
+use function reset;
+
 /**
  * Route defines a mapping from URL to callback / name and vice versa
  */
 final class Route implements MiddlewareInterface, RequestHandlerInterface
 {
+    private string $name;
```
samdark: We need to initialize it since it's optional.

```suggestion
    private ?string $name = null;
```

src/Route.php
```diff
@@ -2,56 +2,36 @@
...
+    /** @var string[] */
+    private array $methods;
```
samdark: In `public function __toString()` it checks for `null` which looks incorrect to me. Should check for empty array instead.

src/RouterFactory.php
```diff
@@ -2,14 +2,26 @@
...
+     */
     private $engineFactory;
-    private $routes;
+    /**
+     * @var array
```
samdark: No need to duplicated type in phpdoc.

src/MatchingResult.php
viktorprogger: It's true for now, but what will be after a year of changes?

src/Group.php
viktorprogger: I'm not agree with you. I can extend this class in my project and put given parameters (including default values) to the parent:
```php
    public function __construct(?string $prefix = null, ?callable $callback = null)
    {
        // my stuff
        parent::__construct($prefix, $callback);
    }
```

src/MatchingResult.php
samdark: We'll see in a year. What we need to think upfront are interfaces, not implementation detail. It should solve a problem and just that.

src/Group.php
samdark: OK. That's valid.

samdark:
Looks OK but it seems travis isn't happy :)

src/RouterFactory.php
```diff
@@ -2,14 +2,26 @@
...
+
 final class RouterFactory
 {
+    /**
+     * @var callable
+     */
     private $engineFactory;
-    private $routes;
+    /**
+     * @var Route[]|array
```
roxblnfk: I think here you can leave only `Route[]`

viktorprogger:
Sorry, I was in a hurry yesterday. Going fix everything today.

samdark:
Thanks!

PR #19, opened 2020-01-30
Router fix

yiiliveext:
| Q             | A
| ------------- | ---
| Is bugfix?    | ✔️
| New feature?  |❌
| Breaks BC?    | ❌
| Tests pass?   | ✔️

Fix codestyle and Route handler logical error

### DIFF
--- a/src/Middleware/Callback.php
+++ b/src/Middleware/Callback.php
@@ -1,6 +1,5 @@
 <?php
 
-
 namespace Yiisoft\Router\Middleware;
 
 use Psr\Http\Message\ResponseInterface;
--- a/src/Route.php
+++ b/src/Route.php
@@ -261,7 +261,7 @@ final class Route implements MiddlewareInterface, RequestHandlerInterface
         $middleware = current($this->middlewares);
         next($this->middlewares);
         if ($middleware === false) {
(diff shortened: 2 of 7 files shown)


yiiliveext:
It is strange. According to PSR-12 (https://www.php-fig.org/psr/psr-12/#8-anonymous-classes) 
```php 
new class implements MiddlewareInterface {
}
```
is the right way. But continuous-integration/styleci/pr thinks otherwise.

roxblnfk:
> This is strange. According to PSR-12 (https://www.php-fig.org/psr/psr-12/#8-anonymous-classes)
> 
> ```
> new class implements MiddlewareInterface {
> }
> ```
> 
> is the right way. But continuous-integration/styleci/pr thinks otherwise.

psr:
`new class implements MiddlewareInterface`
styleci:
`new class() implements MiddlewareInterface`
linter:
`new class implements MiddlewareInterface` or `new class () implements MiddlewareInterface`

( ╯°□°)╯┻┻

But PSR could be adjusted soon: https://github.com/php-fig/fig-standards/pull/1206

samdark:
Good catch about the error.

samdark:
Thanks!

PR #20, opened 2020-01-31
Router alt nested route

yiiliveext:
| Q             | A
| ------------- | ---
| Is bugfix?    | ✔️
| New feature?  | ❌
| Breaks BC?    | ❌
| Tests pass?   | ✔️

This is a little bit better implementation of the Route class.

There are a few significant disadvantages of the current implementation.
1. It is the ability of jumping over handlers.
Example
```php
//Route
Route::get('/test')
                ->to(new ActionCaller(SiteController::class, 'test', $container))
                ->prepend(function (ServerRequestInterface $request, RequestHandlerInterface $handler) use ($container) {
                    $response = $handler->handle($request);
                    $response->getBody()->write(' Sweet');
                    return $response;
                })
                ->prepend(function (ServerRequestInterface $request, RequestHandlerInterface $handler) use ($container) {
                    $response = $handler->handle($request);
                    $response->getBody()->write(' Home');
                    return $response;
                })
                ->name('site/test'),
//Action
public function test(RouterInterface $router, ServerRequestInterface $request, RequestHandlerInterface $handler): ResponseInterface
    {
        $response = $this->responseFactory->createResponse();
        $response->getBody()->write('My');

        return $response;
    }
```
Everything is ok now. The output for both implementations is
```
My Sweet Home
```
What about breaking it? Let's try it!
```php
Route::get('/test')
                ->to(new ActionCaller(SiteController::class, 'test', $container))
                ->prepend(function (ServerRequestInterface $request, RequestHandlerInterface $handler) use ($container) {
                    $response = $handler->handle($request);
                    $response->getBody()->write(' Sweet');
                    return $response;
                })
                ->prepend(function (ServerRequestInterface $request, RequestHandlerInterface $handler) use ($container) {
                    $handler->handle($request);
                    $response = $handler->handle($request);
                    $response->getBody()->write(' Home');
                    return $response;
                })
                ->name('site/test'),
``` 
The output for the current inplementation is
```
We were unable to find the page /test. Home
```
The output for my implementation is
```
My Sweet Home
```

```php
Route::get('/test')
                ->to(new ActionCaller(SiteController::class, 'test', $container))
                ->prepend(function (ServerRequestInterface $request, RequestHandlerInterface $handler) use ($container) {
                    $handler->handle($request);
                    $response = $handler->handle($request);
                    $response->getBody()->write(' Sweet');
                    return $response;
                })
                ->prepend(function (ServerRequestInterface $request, RequestHandlerInterface $handler) use ($container) {
                    $response = $handler->handle($request);
                    $response->getBody()->write(' Home');
                    return $response;
                })
                ->name('site/test'),
``` 
The output for the current inplementation is
```
We were unable to find the page /test. Sweet Home
```
The output for my implementation is
```
My Sweet Home
```

2.  Ability to get an infinite loop.
```php
Route::get('/test')
                ->to(new ActionCaller(SiteController::class, 'test', $container))
                ->prepend(function (ServerRequestInterface $request, RequestHandlerInterface $handler) use ($container) {
                    $handler->dispath($request);
                    $response = $handler->handle($request);
                    $response->getBody()->write(' Sweet');
                    return $response;
                })
                ->prepend(function (ServerRequestInterface $request, RequestHandlerInterface $handler) use ($container) {
                    $response = $handler->handle($request);
                    $response->getBody()->write(' Home');
                    return $response;
                })
                ->name('site/test'),
```
The result for the current inplementation is
```
Infinite loop, HTTP ERROR 500
```
My implementation doesn't have the dispatch() method.

```php
Route::get('/test')
                ->to(new ActionCaller(SiteController::class, 'test', $container))
                ->prepend(function (ServerRequestInterface $request, RequestHandlerInterface $handler) use ($container) {
                    $handler->process($request, $handle);
                    $response = $handler->handle($request);
                    $response->getBody()->write(' Sweet');
                    return $response;
                })
                ->prepend(function (ServerRequestInterface $request, RequestHandlerInterface $handler) use ($container) {
                    $response = $handler->handle($request);
                    $response->getBody()->write(' Home');
                    return $response;
                })
                ->name('site/test'),
```
The result for the current inplementation is
```
Infinite loop, HTTP ERROR 500
```
My implementation doesn't have the process() method.

Basically the only problem is that we don't have the Route object (as handler) in each middleware.
But I've created a fix for that in the router-fastroute package https://github.com/yiiliveext/router-fastroute/commit/93849001f82d2111b3a7fce91d7ca5cfbb1cbf7a. This is even better. We have the ability to get the current Route object from anywhere we can get RouterInterface::class from the container.

### DIFF
--- a/src/Route.php
+++ b/src/Route.php
@@ -28,6 +28,9 @@ final class Route implements MiddlewareInterface
     private array $middlewares = [];
     private array $defaults = [];
 
+    private function __construct()
+    {
+    }
 
     public static function get(string $pattern): self
     {
@@ -266,8 +269,8 @@ final class Route implements MiddlewareInterface
     private function wrap(MiddlewareInterface $middleware, RequestHandlerInterface $handler): RequestHandlerInterface
     {
         return new class($middleware, $handler) implements RequestHandlerInterface {
-            private $middleware;
-            private $handler;
+            private MiddlewareInterface $middleware;
+            private RequestHandlerInterface $handler;
 
             public function __construct(MiddlewareInterface $middleware, RequestHandlerInterface $handler)
             {
--- a/src/Route.php
+++ b/src/Route.php
@@ -39,79 +39,37 @@ final class Route implements MiddlewareInterface
 
     public static function get(string $pattern, $middleware = null): self
     {
-        $route = new static();
-        $route->methods = [Method::GET];
-        $route->pattern = $pattern;
-        if ($middleware !== null) {
-            $route->middlewares[] = $route->prepareMiddleware($middleware);
-        }
-        return $route;
+        return static::methods([Method::GET], $pattern, $middleware);
     }
 
     public static function post(string $pattern, $middleware = null): self
     {
-        $route = new static();
-        $route->methods = [Method::POST];
-        $route->pattern = $pattern;
-        if ($middleware !== null) {
-            $route->middlewares[] = $route->prepareMiddleware($middleware);
-        }
-        return $route;
+        return static::methods([Method::POST], $pattern, $middleware);
     }
 
     public static function put(string $pattern, $middleware = null): self
     {
-        $route = new static();
-        $route->methods = [Method::PUT];
-        $route->pattern = $pattern;
-        if ($middleware !== null) {
-            $route->middlewares[] = $route->prepareMiddleware($middleware);
-        }
-        return $route;
+        return static::methods([Method::PUT], $pattern, $middleware);
     }
 
     public static function delete(string $pattern, $middleware = null): self
     {
-        $route = new static();
-        $route->methods = [Method::DELETE];
-        $route->pattern = $pattern;
-        if ($middleware !== null) {
-            $route->middlewares[] = $route->prepareMiddleware($middleware);
-        }
-        return $route;
+        return static::methods([Method::DELETE], $pattern, $middleware);
     }
 
     public static function patch(string $pattern, $middleware = null): self
     {
-        $route = new static();
-        $route->methods = [Method::PATCH];
-        $route->pattern = $pattern;
-        if ($middleware !== null) {
-            $route->middlewares[] = $route->prepareMiddleware($middleware);
-        }
-        return $route;
+        return static::methods([Method::PATCH], $pattern, $middleware);
     }
 
     public static function head(string $pattern, $middleware = null): self
     {
-        $route = new static();
-        $route->methods = [Method::HEAD];
-        $route->pattern = $pattern;
-        if ($middleware !== null) {
-            $route->middlewares[] = $route->prepareMiddleware($middleware);
-        }
-        return $route;
+        return static::methods([Method::HEAD], $pattern, $middleware);
     }
 
     public static function options(string $pattern, $middleware = null): self
     {
-        $route = new static();
-        $route->methods = [Method::OPTIONS];
-        $route->pattern = $pattern;
-        if ($middleware !== null) {
-            $route->middlewares[] = $route->prepareMiddleware($middleware);
-        }
-        return $route;
+        return static::methods([Method::OPTIONS], $pattern, $middleware);
     }
 
     public static function methods(array $methods, string $pattern, $middleware = null): self
@@ -127,13 +85,7 @@ final class Route implements MiddlewareInterface
 
     public static function anyMethod(string $pattern, $middleware = null): self
     {
-        $route = new static();
-        $route->methods = Method::ANY;
-        $route->pattern = $pattern;
-        if ($middleware !== null) {
-            $route->middlewares[] = $route->prepareMiddleware($middleware);
-        }
-        return $route;
+        return static::methods(Method::ANY, $pattern, $middleware);
     }
 
     public function name(string $name): self
--- a/src/Route.php
+++ b/src/Route.php
@@ -3,33 +3,33 @@
 namespace Yiisoft\Router;
 
 use InvalidArgumentException;
-use LogicException;
 use Yiisoft\Http\Method;
 use Yiisoft\Router\Middleware\Callback;
-use Psr\Http\Message\ResponseInterface;
-use Psr\Http\Message\ServerRequestInterface;
 use Psr\Http\Server\MiddlewareInterface;
 use Psr\Http\Server\RequestHandlerInterface;
+use Psr\Http\Message\ServerRequestInterface;
+use Psr\Http\Message\ResponseInterface;
 
 /**
  * Route defines a mapping from URL to callback / name and vice versa
  */
-final class Route implements MiddlewareInterface, RequestHandlerInterface
+final class Route implements MiddlewareInterface
 {
     private ?string $name = null;
     /** @var string[] */
     private array $methods;
     private string $pattern;
     private ?string $host = null;
+
     /**
      * @var MiddlewareInterface[]|callable[]
      */
-    private array $middlewares = [];
     private array $defaults = [];
-    private RequestHandlerInterface $nextHandler;
+    protected array $middlewares = [];
 
-    private function __construct()
+    public function __construct(MiddlewareInterface ...$middlewares)
     {
+        $this->middlewares = $middlewares;
     }
 
     public static function get(string $pattern): self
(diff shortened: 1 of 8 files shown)


src/Route.php
```diff
@@ -3,34 +3,31 @@
...
  */
-final class Route implements MiddlewareInterface, RequestHandlerInterface
+final class Route implements MiddlewareInterface
 {
     private ?string $name = null;
     /** @var string[] */
     private array $methods;
     private string $pattern;
     private ?string $host = null;
+    private ?RequestHandlerInterface $stack = null;
```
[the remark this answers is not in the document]
samdark: Is it possible to do it without wrapping same way as https://github.com/yiisoft/yii-web/blob/master/src/MiddlewareDispatcher.php?

src/Route.php
```diff
@@ -251,35 +248,37 @@ public function getDefaults(): array
...
+    /**
+     * Wraps handler by middlewares
+     */
+    private function wrap(MiddlewareInterface $middleware, RequestHandlerInterface $handler): RequestHandlerInterface
     {
-        reset($this->middlewares);
-        return $this->handle($request);
-    }
+        return new class($middleware, $handler) implements RequestHandlerInterface {
+            private $middleware;
```
samdark:
```suggestion
            private MiddlewareInterface $middleware;
```

src/Route.php
```diff
@@ -251,35 +248,37 @@ public function getDefaults(): array
...
+            private $handler;
```
samdark:
```suggestion
            private RequestHandlerInterface $handler;
```

src/RouterFactory.php
```diff
@@ -26,7 +26,7 @@ public function __invoke(ContainerInterface $container): RouterInterface
     {
         $factory = $this->engineFactory;
         /* @var $router RouterInterface */
-        $router = $factory($container);
+        $router = $factory();
```
samdark: What's the reasoning to remove container from callable factory here?

tests/RouteTest.php
```diff
@@ -169,46 +170,155 @@ static function (): ResponseInterface {
         $this->assertSame(418, $response->getStatusCode());
     }
 
-    public function testThen(): void
+    public function testThenByStatusCode(): void
```
samdark: There are two tests in this test-method:

1. Checking that whole middleware stack for a route is processed.
2. Checking that middleware stack execution could be interrupted.

Worth splitting these into two separate test methods.

tests/RouteTest.php
```diff
@@ -169,46 +170,155 @@ static function (): ResponseInterface {
...
 
-        $route->process($request, $this->getRequestHandler());
+        $routeTwo = $routeTwo->to($middleware1)->then($middleware2);
+
+        $response = $routeTwo->process($request, $this->getRequestHandler());
+        $this->assertSame(404, $response->getStatusCode());
     }
 
-    public function testBefore(): void
+    public function testThenByBody(): void
```
samdark: This test is similar to `testThenByStatusCode`. It has two tests and these are checking same things: execution of the whole middleware stack and ability to stop stack processing. Is there a need for duplicating the test?

tests/RouteTest.php
```diff
@@ -169,46 +170,155 @@ static function (): ResponseInterface {
...
+
+        $routeTwo = $routeTwo->to($middleware1)->then($middleware2);
+
+        $response = $routeTwo->process($request, $this->getRequestHandler());
+        $body = $response->getBody();
+        $body->rewind();
+        $this->assertSame('middleware1', $response->getBody()->getContents());
+    }
+
+    public function testBeforeByStatusCode(): void
```
samdark: There are two tests in the method. Worth splitting.

tests/RouteTest.php
```diff
@@ -169,46 +170,155 @@ static function (): ResponseInterface {
...
+            return new Response(200);
+        });
+
+        $routeTwo = $routeTwo->to($middleware1)->then($middleware2);
+
+        $response = $routeTwo->process($request, $this->getRequestHandler());
+        $this->assertSame(404, $response->getStatusCode());
+    }
+
+    public function testBeforeByBody(): void
```
samdark: This one is duplicating `testBeforeByStatusCode` by what it is testing.

samdark:
Overall 👍 but there are some important points to consider.

src/RouterFactory.php
yiiliveext: > What's the reasoning to remove container from callable factory here?

It not used here 
https://github.com/yiisoft/router-fastroute/blob/master/src/FastRouteFactory.php

src/Route.php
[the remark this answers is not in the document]
yiiliveext: > Is it possible to do it without wrapping same way as https://github.com/yiisoft/yii-web/blob/master/src/MiddlewareDispatcher.php?
Yes, but the current implementation also has some advantages.
After this PR the handler in the last called middleware will always be an instance of the MiddlewareDispatcher. If a controller action will be the last called middleware as we discussed here https://github.com/yiisoft/yii-web/pull/210, we can add some fallback/additional middleware dynamically after RouterMiddleware (before a NotFoundHandler).

src/Route.php
```diff
@@ -3,34 +3,31 @@
...
+    private ?RequestHandlerInterface $stack = null;
+
     /**
      * @var MiddlewareInterface[]|callable[]
      */
     private array $middlewares = [];
     private array $defaults = [];
-    private RequestHandlerInterface $nextHandler;
 
-    private function __construct()
```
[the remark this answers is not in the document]
yiiliveext: Sorry, I got excited.

tests/RouteTest.php
yiiliveext: No need, just historically it was done during the development process. Which of them will we leave?

src/Route.php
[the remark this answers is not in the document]
yiiliveext: https://github.com/yiisoft/router/pull/20/commits/73a422ad551a6cd137e2314fb32307d8620172cc

tests/RouteTest.php
yiiliveext: I agree, and have a question https://github.com/yiisoft/router/pull/20#discussion_r373853660

src/Route.php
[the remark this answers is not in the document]
yiiliveext: > I think it worth adding a phpdoc here describing that the stack holds middleware already wrapped into each other. Currently, based on the name one can assume it's just an array of middleware.

Done.

src/RouterFactory.php
samdark: OK.

tests/RouteTest.php
samdark: Statuses, I think. They're a bit easier to read.

src/Route.php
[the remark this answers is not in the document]
samdark: > After this PR the handler in the last called middleware will always be an instance of the MiddlewareDispatcher. 

OK. Is it reflected in tests? It seems to be important design thing.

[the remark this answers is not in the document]
yiiliveext: > OK. Is it reflected in tests? It seems to be important design thing.
No, at first we need to deсide about the signature of the Route get(put, post etc) method and methods for adding a middleware.

tests/RouteTest.php
yiiliveext: > Statuses, I think. They're a bit easier to read.

I agree, moreover it is faster.

src/Route.php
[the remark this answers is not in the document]
samdark: 1. A route without middleware may make sense (it could be for creating URLs only) so changing factory methods signature isn't needed.
2. A single method for adding (prepending) a middleware is enough. Choose a name you like.

tests/RouteTest.php
yiiliveext: Done.

tests/RouteTest.php
yiiliveext: Done.

src/Route.php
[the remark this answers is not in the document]
yiiliveext: 1. It could be like `get(string $pattern, ?MiddlewareInterface = null)`
It is very convenient to set a controller action in the method directly. And very clear.
If you aren't using additional middlewares you can just write 
```php
Route::get('/', new ActionCaller(SiteController::class, 'index', $container));
```
And it is a warranty that the controller action will be processed last. It is important for begginers.
2. I like `addMiddleware(MiddlewareInterface $middleware)`.

Do you approve of removing the to(), then() methods and renaming the prepend() method to addMiddleware()?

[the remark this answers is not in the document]
yiiliveext: Check the last version:)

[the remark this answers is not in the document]
samdark: 1. OK. Makes sense.
2. That's fine. I approve removal and renaming.

samdark:
Thanks!

PR #21, opened 2020-02-03
Adjust group implementation

yiiliveext:
| Q             | A
| ------------- | ---
| Is bugfix?    | ✔️
| New feature?  | ✔️
| Breaks BC?    | ❌
| Tests pass?   | ✔️

This is a little bit better implementation of the Group class.
I had to add router-fastroute as a dev-dependency, but I don't like it.
I don't like that one part of the Group middleware implementation is in the Group class while another part is in the FastRoute class. It seems to me this is a very bad idea. It should be implemented inside the Group class completely. What do you think?

And now is the time to talk about pleasant. We have a new alternative group syntaxis. It is simpler and more flexible. 

The current syntax
```php
$routes = [
    Route::get('/', new ActionCaller(SiteController::class, 'index', $container)),
    ['/blog', function (Group $group) use ($container, $middleware1, $middleware2) {
        $group->addRoute(Route::get('/login', new ActionCaller(BlogController::class, 'login', $container)));
        $group->addGroup('/post', function (Group $group) use ($container) {
            $group->addRoute(Route::get('/index', new ActionCaller(PostController::class, 'index', $container)));
            $group->addRoute(Route::get('/view', new ActionCaller(PostController::class, 'view', $container)));
        });
        $group->addMiddleware($middleware1)->addMiddleware($middleware2);
    }],
]
```
the new alternative syntax
```php
$routes = [
    Route::get('/', new ActionCaller(SiteController::class, 'index', $container)),
    Group::create('/blog', [
        Route::get('/login', new ActionCaller(BlogController::class, 'login', $container)),
        Group::create('/post', [
            Route::get('/index', new ActionCaller(PostController::class, 'index', $container)),
            Route::get('/view', new ActionCaller(PostController::class, 'view', $container)),
         ])
    ])->addMiddleware($middleware1)->addMiddleware($middleware2),
]
```
I'm working on a yii3 module system now and that's what I need, the flexible way to build routes dinamically.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -27,7 +27,8 @@
     "require-dev": {
         "phpunit/phpunit": "^8.4",
         "psr/http-factory": "^1.0",
-        "nyholm/psr7": "^1.0"
+        "nyholm/psr7": "^1.0",
+        "yiisoft/router-fastroute": "^3.0@dev"
     },
     "autoload": {
         "psr-4": {
--- a/src/Group.php
+++ b/src/Group.php
@@ -21,6 +21,13 @@ class Group implements RouteCollectorInterface
         }
     }
 
+    final public static function create(?string $prefix, array $routes = [])
+    {
+        $factory = new GroupFactory();
+
+        return $factory($prefix, $routes);
+    }
+
     final public function addRoute(Route $route): void
     {
         $this->items[] = $route;
@@ -29,6 +36,11 @@ class Group implements RouteCollectorInterface
     final public function addGroup(string $prefix, callable $callback): void
     {
         $this->items[] = new Group($prefix, $callback);
+    }
+
+    final public function addGroupInstance(Group $group): void
+    {
+        $this->items[] = $group;
     }
 
     /**
@@ -45,7 +57,7 @@ class Group implements RouteCollectorInterface
             throw new InvalidArgumentException('Parameter should be either a PSR middleware or a callable.');
         }
 
-        $this->middlewares[] = $middleware;
+        array_unshift($this->middlewares, $middleware);
 
         return $this;
     }
--- a/src/GroupFactory.php
+++ b/src/GroupFactory.php
@@ -1,0 +1,27 @@
+<?php
+
+namespace Yiisoft\Router;
+
+use InvalidArgumentException;
+
+final class GroupFactory
+{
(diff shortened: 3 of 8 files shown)


yiiliveext:
Lady and gentlemen, have you nothing to say?

samdark:
Thought you're still force-pushing changes :)

composer.json
```diff
@@ -27,7 +27,8 @@
     "require-dev": {
         "phpunit/phpunit": "^8.4",
         "psr/http-factory": "^1.0",
-        "nyholm/psr7": "^1.0"
+        "nyholm/psr7": "^1.0",
+        "yiisoft/router-fastroute": "^3.0@dev"
```
samdark: Is there a way to avoid using concrete implementation for testing? That doesn't look good at all.

src/Group.php
```diff
@@ -45,7 +57,7 @@ public function __construct(?string $prefix = null, ?callable $callback = null)
             throw new InvalidArgumentException('Parameter should be either a PSR middleware or a callable.');
         }
 
-        $this->middlewares[] = $middleware;
+        array_unshift($this->middlewares, $middleware);
```
samdark: Good catch 👍

yiiliveext:
> That is because FastRoute has [native groups support](https://github.com/nikic/FastRoute#route-groups). For groups it is optimizing regex used and we don't want to lose that.

This is not about that. It is about adding group middlewares inside the FastRoute class. We can do it inside the Group class.

samdark:
> This is not about that. It is about adding group middlewares inside the FastRoute class. We can do it inside the Group class.

Can we handle it in a separate pull request?

yiiliveext:
> Thought you're still force-pushing changes :)

These were not significant changes :stuck_out_tongue_winking_eye:

composer.json
yiiliveext: > Is there a way to avoid using concrete implementation for testing? That doesn't look good at all.

If we remove the adding of group middlewares from the FastRoute class, this package will not be needed.

yiiliveext:
> > This is not about that. It is about adding group middlewares inside the FastRoute class. We can do it inside the Group class.
> 
> Can we handle it in a separate pull request?

Of course.

samdark:
Let's have this one in.

PR #24, opened 2020-02-04
Add middleware lazy loading

yiiliveext:
| Q             | A
| ------------- | ---
| Is bugfix?    | ❌
| New feature?  | ✔️
| Breaks BC?    | ❌
| Tests pass?   | ✔️

Ladies and gentlemen, I'm pleased to present you the lazy loading of middlewares.
For now and forever we have a very fast and light router. All middlewares (controller action is included) can be loaded directly before the Route middleware stack will be executed. It means that the middlewares attached this way will be created after the Router matches the route.

Examples of usage:
```php
$routes = [
   Route::get('/', [SiteController::class, 'index'], $container),
       ->addMiddleware(SomeMiddleware::class),
   Route::get('export', ExportMiddleware::class, $container),
];
```

### DIFF
--- a/src/Middleware/ActionCaller.php
+++ b/src/Middleware/ActionCaller.php
@@ -17,8 +17,8 @@ use Yiisoft\Injector\Injector;
  */
 final class ActionCaller implements MiddlewareInterface
 {
-    private $class;
-    private $method;
+    private string $class;
+    private string $method;
     private ContainerInterface $container;
 
     public function __construct(string $class, string $method, ContainerInterface $container)
--- a/src/Middleware/ActionCaller.php
+++ b/src/Middleware/ActionCaller.php
@@ -1,0 +1,38 @@
+<?php
+
+namespace Yiisoft\Router\Middleware;
+
+use Psr\Container\ContainerInterface;
+use Psr\Http\Message\ResponseInterface;
+use Psr\Http\Message\ServerRequestInterface;
+use Psr\Http\Server\MiddlewareInterface;
+use Psr\Http\Server\RequestHandlerInterface;
+use Yiisoft\Injector\Injector;
+
+/**
+ * ActionCaller maps a route to specified class instance and method.
+ *
+ * Dependencies are automatically injected into both method
+ * and constructor based on types specified.
+ */
+final class ActionCaller implements MiddlewareInterface
+{
+    private $class;
+    private $method;
+    private $container;
+
+    public function __construct(string $class, string $method, ContainerInterface $container)
+    {
+        $this->class = $class;
+        $this->method = $method;
+        $this->container = $container;
+    }
+
+    public function process(ServerRequestInterface $request, RequestHandlerInterface $handler): ResponseInterface
+    {
+        $controller = $this->container->get($this->class);
+        $response = (new Injector($this->container))->invoke([$controller, $this->method], [$request, $handler]);
+        $handler->handle($request);
+        return $response;
+    }
+}
--- a/src/Middleware/ActionCaller.php
+++ b/src/Middleware/ActionCaller.php
@@ -31,8 +31,6 @@ final class ActionCaller implements MiddlewareInterface
     public function process(ServerRequestInterface $request, RequestHandlerInterface $handler): ResponseInterface
     {
         $controller = $this->container->get($this->class);
-        $response = (new Injector($this->container))->invoke([$controller, $this->method], [$request, $handler]);
-        $handler->handle($request);
-        return $response;
+        return (new Injector($this->container))->invoke([$controller, $this->method], [$request, $handler]);
     }
 }
--- a/src/Middleware/ActionCaller.php
+++ b/src/Middleware/ActionCaller.php
@@ -19,7 +19,7 @@ final class ActionCaller implements MiddlewareInterface
 {
     private $class;
     private $method;
-    private $container;
+    private ContainerInterface $container;
 
     public function __construct(string $class, string $method, ContainerInterface $container)
     {
--- a/src/Route.php
+++ b/src/Route.php
@@ -229,7 +229,7 @@ final class Route implements MiddlewareInterface
             return $this->container->get($middleware);
         }
 
-        if (is_array($middleware)) {
+        if (is_array($middleware) && !is_object($middleware[0])) {
             return new ActionCaller($middleware[0], $middleware[1], $this->container);
         }
 
--- a/src/Route.php
+++ b/src/Route.php
@@ -41,6 +41,18 @@ final class Route implements MiddlewareInterface
         $this->container = $container;
     }
 
+    public function setContainer(ContainerInterface $container)
+    {
+        $route = clone $this;
+        $route->container = $container;
+        return $route;
+    }
+
+    public function hasContainer()
+    {
+        return $this->container !== null;
+    }
+
     /**
      * @param string $pattern
      * @param MiddlewareInterface|callable|string|array|null $middleware primary route handler {@see addMiddleware()}
@@ -190,9 +202,6 @@ final class Route implements MiddlewareInterface
         if (
             is_string($middleware) && is_subclass_of($middleware, MiddlewareInterface::class)
         ) {
-            if ($this->container === null) {
-                throw new InvalidArgumentException('Route container must not be null for lazy loaded middleware.');
-            }
             return;
         }
 
@@ -201,16 +210,10 @@ final class Route implements MiddlewareInterface
             && isset($middleware[0], $middleware[1])
             && is_string($middleware[1]) && is_string($middleware[0]) && class_exists($middleware[0])
         ) {
-            if ($this->container === null) {
-                throw new InvalidArgumentException('Route container must not be null for handler action.');
-            }
             return;
         }
 
         if (is_callable($middleware)) {
-            if ($this->container === null) {
-                throw new InvalidArgumentException('Route container must not be null for callable.');
-            }
             return;
         }
 
@@ -226,14 +229,23 @@ final class Route implements MiddlewareInterface
     private function prepareMiddleware($middleware)
     {
         if (is_string($middleware)) {
+            if ($this->container === null) {
+                throw new InvalidArgumentException('Route container must not be null for lazy loaded middleware.');
+            }
             return $this->container->get($middleware);
         }
 
         if (is_array($middleware) && !is_object($middleware[0])) {
+            if ($this->container === null) {
+                throw new InvalidArgumentException('Route container must not be null for handler action.');
+            }
             return new ActionCaller($middleware[0], $middleware[1], $this->container);
         }
 
         if (is_callable($middleware)) {
+            if ($this->container === null) {
+                throw new InvalidArgumentException('Route container must not be null for callable.');
+            }
             return new Callback($middleware, $this->container);
         }
 
--- a/src/Route.php
+++ b/src/Route.php
@@ -43,8 +43,8 @@ final class Route implements MiddlewareInterface
 
     /**
      * @param string $pattern
-     * @param callable|MiddlewareInterface|null $middleware
-     * @param ContainerInterface $container
+     * @param MiddlewareInterface|callable|string|array|null $middleware primary route handler {@see addMiddleware()}
+     * @param ContainerInterface $container|null
      * @return static
      */
     public static function get(string $pattern, $middleware = null, ?ContainerInterface $container = null): self
@@ -54,8 +54,8 @@ final class Route implements MiddlewareInterface
 
     /**
      * @param string $pattern
-     * @param callable|MiddlewareInterface|null $middleware
-     * @param ContainerInterface $container
+     * @param MiddlewareInterface|callable|string|array|null $middleware primary route handler {@see addMiddleware()}
+     * @param ContainerInterface|null $container
      * @return static
      */
     public static function post(string $pattern, $middleware = null, ?ContainerInterface $container = null): self
@@ -65,7 +65,8 @@ final class Route implements MiddlewareInterface
 
     /**
      * @param string $pattern
-     * @param callable|MiddlewareInterface|null $middleware
+     * @param MiddlewareInterface|callable|string|array|null $middleware primary route handler {@see addMiddleware()}
+     * @param ContainerInterface|null $container
      * @return static
      */
     public static function put(string $pattern, $middleware = null, ?ContainerInterface $container = null): self
@@ -75,8 +76,8 @@ final class Route implements MiddlewareInterface
 
     /**
      * @param string $pattern
-     * @param callable|MiddlewareInterface|null $middleware
-     * @param ContainerInterface $container
+     * @param MiddlewareInterface|callable|string|array|null $middleware primary route handler {@see addMiddleware()}
+     * @param ContainerInterface|null $container
      * @return static
      */
     public static function delete(string $pattern, $middleware = null, ?ContainerInterface $container = null): self
@@ -86,8 +87,8 @@ final class Route implements MiddlewareInterface
 
     /**
      * @param string $pattern
-     * @param callable|MiddlewareInterface|null $middleware
-     * @param ContainerInterface $container
+     * @param MiddlewareInterface|callable|string|array|null $middleware primary route handler {@see addMiddleware()}
+     * @param ContainerInterface|null $container
      * @return static
      */
     public static function patch(string $pattern, $middleware = null, ?ContainerInterface $container = null): self
@@ -97,8 +98,8 @@ final class Route implements MiddlewareInterface
 
     /**
      * @param string $pattern
-     * @param callable|MiddlewareInterface|null $middleware
-     * @param ContainerInterface $container
+     * @param MiddlewareInterface|callable|string|array|null $middleware primary route handler {@see addMiddleware()}
+     * @param ContainerInterface|null $container
      * @return static
      */
     public static function head(string $pattern, $middleware = null, ?ContainerInterface $container = null): self
@@ -108,8 +109,8 @@ final class Route implements MiddlewareInterface
 
     /**
      * @param string $pattern
-     * @param callable|MiddlewareInterface|null $middleware
-     * @param ContainerInterface $container
+     * @param MiddlewareInterface|callable|string|array|null $middleware primary route handler {@see addMiddleware()}
+     * @param ContainerInterface|null $container
      * @return static
      */
     public static function options(string $pattern, $middleware = null, ?ContainerInterface $container = null): self
@@ -120,8 +121,8 @@ final class Route implements MiddlewareInterface
     /**
      * @param array $methods
      * @param string $pattern
-     * @param callable|MiddlewareInterface|null $middleware
-     * @param ContainerInterface $container
+     * @param MiddlewareInterface|callable|string|array|null $middleware primary route handler {@see addMiddleware()}
+     * @param ContainerInterface|null $container
      * @return static
      */
     public static function methods(array $methods, string $pattern, $middleware = null, ?ContainerInterface $container = null): self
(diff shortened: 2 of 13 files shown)


src/Route.php
```diff
@@ -170,10 +174,31 @@ public function defaults(array $defaults): self
...
      * @param callable|MiddlewareInterface $middleware
-     * @return MiddlewareInterface
+     * @return MiddlewareInterface|string|array
      */
-    private function prepareMiddleware($middleware): MiddlewareInterface
+    private function prepareMiddleware($middleware)
     {
+        if (
+            is_string($middleware) && class_exists($middleware)
+            && in_array(MiddlewareInterface::class, class_implements($middleware))
```
roxblnfk:
```suggestion
            is_string($middleware) && is_subclass_of($middleware, MiddlewareInterface::class)
```

src/Route.php
```diff
@@ -20,6 +22,7 @@ final class Route implements MiddlewareInterface
     private array $methods;
     private string $pattern;
     private ?string $host = null;
+    private $container;
```
roxblnfk:
```suggestion
    private ?ContainerInterface $container;
```

src/Route.php
```diff
@@ -129,9 +133,9 @@ public static function methods(array $methods, string $pattern, $middleware = nu
      * @param callable|MiddlewareInterface|null $middleware
      * @return static
      */
-    public static function anyMethod(string $pattern, $middleware = null): self
+    public static function anyMethod(string $pattern, $middleware = null, $container = null): self
```
roxblnfk:
```suggestion
    public static function anyMethod(string $pattern, $middleware = null, ?ContainerInterface $container = null): self
```

src/Route.php
```diff
@@ -113,9 +117,9 @@ public static function options(string $pattern, $middleware = null): self
      * @param callable|MiddlewareInterface|null $middleware
      * @return static
      */
-    public static function methods(array $methods, string $pattern, $middleware = null): self
+    public static function methods(array $methods, string $pattern, $middleware = null, $container = null): self
```
roxblnfk:
```suggestion
    public static function methods(array $methods, string $pattern, $middleware = null, ?ContainerInterface $container = null): self
```

src/Route.php
```diff
@@ -28,83 +31,84 @@ final class Route implements MiddlewareInterface
...
+        return static::methods([Method::HEAD], $pattern, $middleware, $container);
     }
 
     /**
      * @param string $pattern
      * @param callable|MiddlewareInterface|null $middleware
      * @return static
      */
-    public static function options(string $pattern, $middleware = null): self
+    public static function options(string $pattern, $middleware = null, $container = null): self
```
roxblnfk:
```suggestion
    public static function options(string $pattern, $middleware = null, ?ContainerInterface  $container = null): self
```

src/Route.php
```diff
@@ -28,83 +31,84 @@ final class Route implements MiddlewareInterface
...
+        return static::methods([Method::PATCH], $pattern, $middleware, $container);
     }
 
     /**
      * @param string $pattern
      * @param callable|MiddlewareInterface|null $middleware
      * @return static
      */
-    public static function head(string $pattern, $middleware = null): self
+    public static function head(string $pattern, $middleware = null, $container = null): self
```
roxblnfk:
```suggestion
    public static function head(string $pattern, $middleware = null, ?ContainerInterface $container = null): self
```

src/Route.php
```diff
@@ -28,83 +31,84 @@ final class Route implements MiddlewareInterface
...
+        return static::methods([Method::DELETE], $pattern, $middleware, $container);
     }
 
     /**
      * @param string $pattern
      * @param callable|MiddlewareInterface|null $middleware
      * @return static
      */
-    public static function patch(string $pattern, $middleware = null): self
+    public static function patch(string $pattern, $middleware = null, $container = null): self
```
roxblnfk:
```suggestion
    public static function patch(string $pattern, $middleware = null, ?ContainerInterface $container = null): self
```

src/Route.php
```diff
@@ -28,83 +31,84 @@ final class Route implements MiddlewareInterface
...
+        return static::methods([Method::PUT], $pattern, $middleware, $container);
     }
 
     /**
      * @param string $pattern
      * @param callable|MiddlewareInterface|null $middleware
      * @return static
      */
-    public static function delete(string $pattern, $middleware = null): self
+    public static function delete(string $pattern, $middleware = null, $container = null): self
```
roxblnfk:
```suggestion
    public static function delete(string $pattern, $middleware = null, ?ContainerInterface $container = null): self
```

src/Route.php
```diff
@@ -28,83 +31,84 @@ final class Route implements MiddlewareInterface
...
+        return static::methods([Method::POST], $pattern, $middleware, $container);
     }
 
     /**
      * @param string $pattern
      * @param callable|MiddlewareInterface|null $middleware
      * @return static
      */
-    public static function put(string $pattern, $middleware = null): self
+    public static function put(string $pattern, $middleware = null, $container = null): self
```
roxblnfk:
```suggestion
    public static function put(string $pattern, $middleware = null, ?ContainerInterface $container = null): self
```

src/Route.php
```diff
@@ -28,83 +31,84 @@ final class Route implements MiddlewareInterface
...
+        return static::methods([Method::GET], $pattern, $middleware, $container);
     }
 
     /**
      * @param string $pattern
      * @param callable|MiddlewareInterface|null $middleware
      * @return static
      */
-    public static function post(string $pattern, $middleware = null): self
+    public static function post(string $pattern, $middleware = null, $container = null): self
```
roxblnfk:
```suggestion
    public static function post(string $pattern, $middleware = null, ?ContainerInterface $container = null): self
```

src/Route.php
```diff
@@ -28,83 +31,84 @@ final class Route implements MiddlewareInterface
...
+        $this->container = $container;
     }
 
     /**
      * @param string $pattern
      * @param callable|MiddlewareInterface|null $middleware
      * @return static
      */
-    public static function get(string $pattern, $middleware = null): self
+    public static function get(string $pattern, $middleware = null, $container = null): self
```
roxblnfk:
```suggestion
    public static function get(string $pattern, $middleware = null, ?ContainerInterface $container = null): self
```

src/Middleware/ActionCaller.php
```diff
@@ -0,0 +1,38 @@
...
+ * ActionCaller maps a route to specified class instance and method.
+ *
+ * Dependencies are automatically injected into both method
+ * and constructor based on types specified.
+ */
+final class ActionCaller implements MiddlewareInterface
+{
+    private $class;
+    private $method;
+    private $container;
```
roxblnfk:
```suggestion
    private string $class;
    private string $method;
    private ContainerInterface $container;
```

src/Route.php
```diff
@@ -170,10 +174,31 @@ public function defaults(array $defaults): self
...
+        ) {
+            if ($this->container === null) {
+                throw new InvalidArgumentException('Route container must not be null for lazy loaded middleware.');
+            }
+            return $middleware;
+        }
+
+        if (
+            is_array($middleware) && count($middleware) === 2
+            && isset($middleware[0]) && isset($middleware[1])
```
roxblnfk: it can be `isset($middleware[0], $middleware[1])` or `array_keys($middleware) === [0, 1]`

src/Route.php
```diff
@@ -170,10 +174,31 @@ public function defaults(array $defaults): self
...
+            && is_string($middleware[1]) && is_string($middleware[0]) && class_exists($middleware[0])
```
rustamwin: Do we need to check if a class exists?

src/Route.php
yiiliveext: Thanks, you're right.

src/Route.php
rustamwin: This check not needed

yiiliveext: What exactly?

rustamwin: I think this check is not needed because the next step is checked for a string

src/Route.php
```diff
@@ -170,10 +174,31 @@ public function defaults(array $defaults): self
...
     /**
      * @param callable|MiddlewareInterface $middleware
-     * @return MiddlewareInterface
+     * @return MiddlewareInterface|string|array
      */
-    private function prepareMiddleware($middleware): MiddlewareInterface
+    private function prepareMiddleware($middleware)
     {
+        if (
+            is_string($middleware) && class_exists($middleware)
```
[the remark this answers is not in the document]
yiiliveext: Yes

src/Route.php
rustamwin: Because the next step is checked for a string
```php
$middleware = [null, 'index']; // false
$middleware = [false, 'index']; // false
$middleware = [true, 'index']; // false
```

roxblnfk: `array_keys($middleware) === [0, 1]` can replace `count($middleware) === 2 && isset($middleware[0]) && isset($middleware[1])`

> because the next step is checked for a string

if i pass array like `['middleware' => SomeClass::class, 'method' => 'index']` then next step will not work correctly

src/Route.php
yiiliveext: Yes, I think that in this way we can get a more informative error message for the developer.

yiiliveext:
I've copied the ActionCaller class from the yii-web package. I think it would be better to remove this class from the yii-web after this PR will be merged.

src/Route.php
yiiliveext: What do you mean exactly?
1. What condition will we remove.
2. What is the next step.

yiiliveext: > `array_keys($middleware) === [0, 1]` can replace `count($middleware) === 2 && isset($middleware[0]) && isset($middleware[1])`

Is it faster?

roxblnfk: > Is it faster?

¯\\\_(ツ)\_/¯

rustamwin: 1. Condition `isset` is not needed
2. I mean condition `is_string`

yiiliveext: No, you're wrong. 
Use case: $middleware = [null, 'index']; 
is_string($middleware[0])  //an exception will be thrown

rustamwin: Oh, you're right. I'm so tired :(

src/Route.php
```diff
@@ -252,6 +276,11 @@ public function process(ServerRequestInterface $request, RequestHandlerInterface
     {
         if ($this->stack === null) {
             for ($i = count($this->middlewares) - 1; $i >= 0; $i--) {
```
xepozz: I think would be nice to change the loop from `for` to `foreach`. What do you think?

tests/RouteTest.php
```diff
@@ -216,6 +219,96 @@ public function testMiddlewareStackInterrupted(): void
         $this->assertSame(403, $response->getStatusCode());
     }
 
+    public function testInvalidMiddlewareAddWrongStringLL()
```
xepozz: What is the `LL` on the ending of method names?

tests/Support/TestMiddleware.php
```diff
@@ -0,0 +1,23 @@
...
+use Nyholm\Psr7\Response;
+use Psr\Http\Server\MiddlewareInterface;
+use Psr\Http\Server\RequestHandlerInterface;
+use Psr\Http\Message\ServerRequestInterface;
+use Psr\Http\Message\ResponseInterface;
+
+/**
+ * TestMiddleware
+ */
+class TestMiddleware implements MiddlewareInterface
```
xepozz: Can be inlined to the test?

tests/Support/TestController.php
```diff
@@ -0,0 +1,20 @@
...
+
+/**
+ * ETestController
+ */
+class TestController
+{
+    /**
+     * @return string
+     */
+    public function index(): ResponseInterface
```
xepozz: Can be inlined to the test?

src/Route.php
```diff
@@ -170,10 +174,30 @@ public function defaults(array $defaults): self
...
+            && is_string($middleware[1]) && is_string($middleware[0]) && class_exists($middleware[0])
+        ) {
+            if ($this->container === null) {
+                throw new InvalidArgumentException('Route container must not be null for lazy loaded middleware.');
+            }
+            return $middleware;
+        }
+
         if (is_callable($middleware)) {
             $middleware = new Callback($middleware);
```
xepozz:
```suggestion
            return new Callback($middleware);
```

src/Route.php
```diff
@@ -170,10 +174,30 @@ public function defaults(array $defaults): self
...
+            is_string($middleware) && is_subclass_of($middleware, MiddlewareInterface::class)
+        ) {
+            if ($this->container === null) {
+                throw new InvalidArgumentException('Route container must not be null for lazy loaded middleware.');
+            }
+            return $middleware;
+        }
+
+        if (
+            is_array($middleware) && count($middleware) === 2
```
xepozz:
```suggestion
            is_array($middleware) && count($middleware) === 2 &&
```

src/Route.php
```diff
@@ -170,10 +174,30 @@ public function defaults(array $defaults): self
...
+            && isset($middleware[0], $middleware[1])
```
xepozz:
```suggestion
```

src/Route.php
```diff
@@ -170,10 +174,30 @@ public function defaults(array $defaults): self
...
+            && is_string($middleware[1]) && is_string($middleware[0]) && class_exists($middleware[0])
```
xepozz:
```suggestion
            is_string(reset($middleware)) && is_string(next($middleware)) && class_exists(reset($middleware))
```
At now we can check any indexes, but I don't like this solution :(

src/Route.php
```diff
@@ -170,10 +174,30 @@ public function defaults(array $defaults): self
...
+        if (
+            is_string($middleware) && is_subclass_of($middleware, MiddlewareInterface::class)
+        ) {
+            if ($this->container === null) {
+                throw new InvalidArgumentException('Route container must not be null for lazy loaded middleware.');
+            }
+            return $middleware;
+        }
+
+        if (
```
xepozz: Can it be replaced with `is_callable` :question:

src/Route.php
```diff
@@ -252,6 +276,11 @@ public function process(ServerRequestInterface $request, RequestHandlerInterface
     {
         if ($this->stack === null) {
             for ($i = count($this->middlewares) - 1; $i >= 0; $i--) {
+                if (is_string($this->middlewares[$i])) {
+                    $this->middlewares[$i] = $this->container->get($this->middlewares[$i]);
+                } elseif (is_array($this->middlewares[$i])) {
```
xepozz: How about to check indexes? :)

tests/RouteTest.php
rustamwin: I think lazy loading

src/Route.php
yiiliveext: See the coversations above

src/Route.php
yiiliveext: See tests

src/Route.php
yiiliveext: It was checked in the prepareMiddleware()

src/Route.php
yiiliveext: No

src/Route.php
yiiliveext: Is it the PSR-12 rule?

src/Route.php
yiiliveext: Are you sure that a call of the array_reverse() is a good idea?

src/Route.php
```diff
@@ -113,9 +117,9 @@ public static function options(string $pattern, $middleware = null): self
      * @param callable|MiddlewareInterface|null $middleware
      * @return static
      */
-    public static function methods(array $methods, string $pattern, $middleware = null): self
+    public static function methods(array $methods, string $pattern, $middleware = null, ?ContainerInterface $container = null): self
```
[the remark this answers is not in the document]
yiiliveext: Ok, I will check it.

src/Middleware/ActionCaller.php
```diff
@@ -0,0 +1,38 @@
...
+        $this->class = $class;
+        $this->method = $method;
+        $this->container = $container;
+    }
+
+    public function process(ServerRequestInterface $request, RequestHandlerInterface $handler): ResponseInterface
+    {
+        $controller = $this->container->get($this->class);
+        $response = (new Injector($this->container))->invoke([$controller, $this->method], [$request, $handler]);
+        $handler->handle($request);
```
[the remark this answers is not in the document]
yiiliveext: It is the same as with https://github.com/yiisoft/yii-web/blob/master/src/Middleware/ActionCaller.php

src/Middleware/ActionCaller.php
```diff
@@ -0,0 +1,38 @@
...
+    {
+        $this->class = $class;
+        $this->method = $method;
+        $this->container = $container;
+    }
+
+    public function process(ServerRequestInterface $request, RequestHandlerInterface $handler): ResponseInterface
+    {
+        $controller = $this->container->get($this->class);
+        $response = (new Injector($this->container))->invoke([$controller, $this->method], [$request, $handler]);
```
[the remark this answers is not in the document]
yiiliveext: Maybe this is a good idea.

yiiliveext:
> Can we avoid the last argument `$container` from routes methods?
> It looks bad :(

How can we do this?

tests/Support/TestMiddleware.php
yiiliveext: No, need a class name

tests/Support/TestController.php
yiiliveext: No, needs a class name

src/Middleware/ActionCaller.php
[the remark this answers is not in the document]
yiiliveext: > Why is this called and not returned from the method?

Thanks, it is not needed after my previous PR.

src/Middleware/ActionCaller.php
```diff
@@ -31,8 +31,6 @@ public function __construct(string $class, string $method, ContainerInterface $c
     public function process(ServerRequestInterface $request, RequestHandlerInterface $handler): ResponseInterface
     {
         $controller = $this->container->get($this->class);
-        $response = (new Injector($this->container))->invoke([$controller, $this->method], [$request, $handler]);
```
yiiliveext: Maybe it is still necessary because we have the ability to add ActionCaller in random order.

src/Route.php
```diff
@@ -169,17 +183,58 @@ public function defaults(array $defaults): self
...
+            && is_string($middleware[1]) && is_string($middleware[0]) && class_exists($middleware[0])
+        ) {
+            if ($this->container === null) {
+                throw new InvalidArgumentException('Route container must not be null for handler action.');
+            }
+            return;
+        }
+
         if (is_callable($middleware)) {
-            $middleware = new Callback($middleware);
```
yiiliveext: The previous check is not necessary now, because is_callable() checks that this is the array with two elements where first is a class name or an object and second it is the class method. But we have to fix the expected error message in the test. What do you think?

src/Route.php
```diff
@@ -169,17 +183,58 @@ public function defaults(array $defaults): self
...
+     * @param MiddlewareInterface|callable|string|array $middleware
+     * @return MiddlewareInterface|string|array
+     */
+    private function prepareMiddleware($middleware)
+    {
+        if (is_string($middleware)) {
+            return $this->container->get($middleware);
+        }
+
+        if (is_array($middleware)) {
```
yiiliveext: A callable can be an array. See my last commit with added test and fix.

xepozz: @yiiliveext `['ClassName', 'methodName']` is a callable too

yiiliveext: > @yiiliveext `['ClassName', 'methodName']` is a callable too

That's what I said.

src/Middleware/ActionCaller.php
samdark: The line was always calling a handler but not using result of the call. If you want to delegate handling to the handler, you can do it in the action code if needed.

src/Route.php
samdark: I think it's good that the messages are specific.

samdark:
Merged. Great changes!

PR #25, opened 2020-02-05
Fix Typed property Yiisoft\Router\Group::$container must not be acces…

thenotsoft:
Typed property `Yiisoft\Router\Group::$container` must not be accessed before
initialization

| Q             | A
| ------------- | ---
| Is bugfix?    | ✔️
| New feature?  | ❌
| Breaks BC?    | ❌
| Tests pass?   | ✔️
| Fixed issues  |


samdark:
Thanks!

PR #26, opened 2020-02-06
Add container auto injection

yiiliveext:
| Q             | A
| ------------- | ---
| Is bugfix?    | ❌
| New feature?  | ✔️
| Breaks BC?    | ✔️
| Tests pass?   | ✔️

Added auto-injection of the container for nested groups and routes. Removed the old syntax for adding groups.


src/Group.php
```diff
@@ -31,6 +31,28 @@ public function __construct(?string $prefix = null, ?callable $callback = null,
...
+                    $item->setContainer($container);
+                }
+                $this->items[$index] = $item;
+            }
+        }
+
+        return $this;
+    }
+
+    public function hasContainer(): bool
```
samdark:
```suggestion
    final public function hasContainer(): bool
```

tests/RouteTest.php
```diff
@@ -300,6 +303,184 @@ public function testMiddlewareCallSuccessArrayWithoutContainerLL(): void
...
+                Route::get('/note/{id}')->name('api-v2-note/view'),
+            ])
+        ];
+
+        $factory = new RouterFactory($this->getEngineFactory(), $routes);
+        $router = $factory($container);
+        $items = $router->getItems();
+
+        $func = function ($item) use (&$func) {
+            $this->assertSame(true, $item->hasContainer());
```
xepozz:
```suggestion
            $this->assertTrue($item->hasContainer());
```

tests/RouteTest.php
```diff
@@ -300,6 +303,184 @@ public function testMiddlewareCallSuccessArrayWithoutContainerLL(): void
...
+                array_walk($items, $func);
+            }
+        };
+        array_walk($items, $func);
+    }
+
+    public function testContainerAutoInjectionByRouterFactory()
+    {
+        $container = $this->getContainer([
+            TestController::class => new TestController(),
```
xepozz: It is useless

src/RouterFactory.php
```diff
@@ -27,18 +27,16 @@ public function __invoke(ContainerInterface $container): RouterInterface
         $factory = $this->engineFactory;
         /* @var $router RouterInterface */
         $router = $factory();
+        if (!$router->hasContainer()) {
+            $router = $router->setContainer($container);
```
xepozz: `RouterInterface` has no `setContainer` and `hasContainer` methods.

src/Group.php
```diff
@@ -31,6 +31,25 @@ public function __construct(?string $prefix = null, ?callable $callback = null,
         return $factory($prefix, $routes);
     }
 
+    final public function setContainer(ContainerInterface $container): Group
+    {
+        $group = clone $this;
+        $group->container = $container;
+        foreach ($group->items as $index => $item) {
+            if (!$item->hasContainer() && $container !== null) {
```
xepozz: Based on argument type `ContainerInterface` it can't be `null`

src/Group.php
```diff
@@ -31,6 +31,25 @@ public function __construct(?string $prefix = null, ?callable $callback = null,
         return $factory($prefix, $routes);
     }
 
+    final public function setContainer(ContainerInterface $container): Group
```
xepozz: I think it must be named as `withContainer` (named strategy for fluent setters).

src/Group.php
```diff
@@ -31,6 +31,25 @@ public function __construct(?string $prefix = null, ?callable $callback = null,
...
+    {
+        $group = clone $this;
```
xepozz: Why it need? Why we can't apply new container value to `$this`?

src/Group.php
```diff
@@ -39,13 +58,11 @@ public function __construct(?string $prefix = null, ?callable $callback = null,
...
 
-    final public function addGroup(string $prefix, callable $callback): void
-    {
-        $this->items[] = new Group($prefix, $callback);
-    }
-
-    final public function addGroupInstance(Group $group): void
+    final public function addGroup(Group $group): void
     {
+        if (!$group->hasContainer() && $this->container !== null) {
```
xepozz:
```suggestion
        if (!$group->hasContainer() && $this->hasContainer()) {
```

src/Group.php
```diff
@@ -31,6 +31,25 @@ public function __construct(?string $prefix = null, ?callable $callback = null,
...
+    }
+
+    final public function hasContainer(): bool
+    {
+        return $this->container !== null;
+    }
+
     final public function addRoute(Route $route): void
     {
         if (!$route->hasContainer() && $this->container !== null) {
```
xepozz:
```suggestion
        if (!$route->hasContainer() && $this->hasContainer()) {
```

tests/RouteTest.php
```diff
@@ -300,6 +303,184 @@ public function testMiddlewareCallSuccessArrayWithoutContainerLL(): void
...
+                Route::get('/note')->name('api-v2-note/index'),
+                Route::get('/note/{id}')->name('api-v2-note/view'),
+            ])
+        ];
+
+        $factory = new RouterFactory($this->getEngineFactory(), $routes);
+        $router = $factory($container);
+        $items = $router->getItems();
+
+        $func = function ($item) use (&$func) {
```
xepozz: This method must be extracted as a private method, because it's used in at least 3 methods.

src/Group.php
samdark: Since it's immutable, naming it with `set` isn't a good idea. Also, `self` should be used as return type since cloned object may be a super-class of `Group`.

```suggestion
    final public function withContainer(ContainerInterface $container): self
```

src/Group.php
samdark: `$container !== null` doesn't make sense since method requires non-null value for `$container`.

src/GroupFactory.php
```diff
@@ -16,16 +16,14 @@ public function __construct(ContainerInterface $container = null)
 
     public function __invoke(?string $prefix = null, array $routes = []): Group
     {
-        $group = new Group($prefix, function (RouteCollectorInterface $collector) use ($routes) {
+        $group = new Group($prefix, function (Group $group) use ($routes) {
```
samdark: Either the change is not correct or an interface should be deleted (there's only a single implementation anyway).

src/RouterFactory.php
```diff
@@ -27,18 +27,16 @@ public function __invoke(ContainerInterface $container): RouterInterface
         $factory = $this->engineFactory;
         /* @var $router RouterInterface */
         $router = $factory();
+        if (!$router->hasContainer()) {
```
samdark: `hasContainer()` and `setContainer()` are not part of `RouterInterface`.

src/Group.php
yiiliveext: agree

src/GroupFactory.php
yiiliveext: you're right

src/RouterFactory.php
yiiliveext: Mmm... what is the decision?

samdark: Could be solved by adding methods to `RouteCollectorInterface`.

src/GroupFactory.php
yiiliveext: > Either the change is not correct or an interface should be deleted (there's only a single implementation anyway).

https://github.com/yiisoft/router/blob/master/src/Group.php#L23

src/Group.php
yiiliveext: Ok

src/Group.php
yiiliveext: Road runner

xepozz: What do you mean? Can you explain more further?

src/Group.php
yiiliveext: Done

src/RouterFactory.php
yiiliveext: Done

tests/RouteTest.php
yiiliveext: Ok

src/Group.php
yiiliveext: WTF?

src/Group.php
yiiliveext: With RR we can set another|empty container dynamically and it will be used in the next request.

yiiliveext:
> New test cases are not relavant for `Route`. It's for `Group` only.
> I suggest you to split tests for `Route` and `Group` and you can see your problem with mocking object.
> I think `RouteCollectorInterface` should have `getRoutes()` and `getGroups()` for this case.

I've created this test before implementation. It is very convenient for catching bugs.
I'm still waiting for one-string-mock.

tests/RouteTest.php
yiiliveext: Done

src/Group.php
xepozz: I don't know case which will use this "feature". All routes will be cached and then dynamical configuring is not be an actual.
Also I don't see tests for this feature.

yiiliveext: It is a bug, not the feature

yiiliveext: > Also I don't see tests for this feature.
This requires an integration test.

xepozz:
> I've created this test before implementation. It is very convenient for catching bugs.
I'm still waiting for one-string-mock.

The main problem is that you use `Group` as part of the implementation of interface `RouterInterface`. Such tests are completely "raw" and related not to the interface, but to a specific implementation (`Group`). It follows that the interface will be unstable.
Rewarding the `Group` with unnecessary methods from `UrlGeneratorInterface` and `UrlMatcherInterface` is not a good idea. In this test, you are testing not `Route` or `Group`. You are testing `RouteFactory`, where you need to verify that the `setContainer()` method is called and that’s it.
As I wrote right away, it’s better to break these huge tests into many small ones, so it turns out in the end.

If you cannot mock the object, the problem is with the object. Either it doesn’t do something, or it doing too much.
I can't write one-line mock, because class or test is done poorly.

yiiliveext:
> I can't write one-line mock, because class or test is done poorly.
Even you remove RouteCollectorIterface interface you can't do it.

samdark:
Let's looks if tests could be improved...

xepozz:
> Even if you remove RouteCollectorIterface interface you can't do it.

Yes, you are rigth and I've described above the reason about it.

yiiliveext:
> Let's looks if tests could be improved...

I think this PR can be merged. I'm working on this package and it is very inconvenient to merge another change. If @xepozz wants he could create another PR that will improve the tests.

yiiliveext:
Only one of the three tests relates to the RouterFactory scope .

samdark:
Going to merge soon.

samdark:
Thanks!

samdark:
@xepozz let's made further improvements separately, OK?

PR #28, opened 2020-02-09
Remove unnecessary condition

yiiliveext:
| Q             | A
| ------------- | ---
| Is bugfix?    | ❌
| New feature?  | ❌
| Breaks BC?    | ❌
| Tests pass?   | ✔️

This check is not necessary now, because is_callable() checks that this is the array with two elements where first is a class name or an object and second is the class method name.


samdark:
Thanks!

PR #31, opened 2020-02-14
Prepare to merge fastroute PR

yiiliveext:
| Q             | A
| ------------- | ---
| Is bugfix?    | ❌
| New feature?  | ✔️
| Breaks BC?    | ❌
| Tests pass?   | ✔️

### DIFF
--- a/src/GroupFactory.php
+++ b/src/GroupFactory.php
@@ -7,25 +7,8 @@ use Psr\Container\ContainerInterface;
 
 final class GroupFactory
 {
-    private ?ContainerInterface $container = null;
-
-    public function __construct(ContainerInterface $container = null)
+    public function __invoke(ContainerInterface $container): Group
     {
-        $this->container = $container;
-    }
-
-    public function __invoke(?string $prefix = null, array $routes = []): Group
-    {
-        return new Group($prefix, static function (Group $group) use ($routes) {
-            foreach ($routes as $route) {
-                if ($route instanceof Route) {
-                    $group->addRoute($route);
-                } elseif ($route instanceof Group) {
-                    $group->addGroup($route);
-                } else {
-                    throw new InvalidArgumentException('Routes should be either instances of Route or Group');
-                }
-            }
-        }, $this->container);
+        return new Group(null, null, $container);
     }
 }
--- a/src/RouteCollection.php
+++ b/src/RouteCollection.php
@@ -8,7 +8,12 @@ use InvalidArgumentException;
 
 final class RouteCollection implements RouteCollectionInterface
 {
-    private $items = [];
+    private ?string $prefix = null;
+
+    /**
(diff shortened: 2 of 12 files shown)


src/GroupFactory.php
```diff
@@ -7,25 +7,8 @@
 
 final class GroupFactory
```
xepozz: Is this class used anywhere?

yiiliveext: Yes, see my the first post of my fastroute PR

src/RouteCollection.php
```diff
@@ -0,0 +1,154 @@
...
+
+    /**
+     * RouteCollection constructor.
+     * @param RouteCollectorInterface $collector
+     * @param string $prefix
+     * @param bool $buildRoutes
+     */
+    public function __construct(RouteCollectorInterface $collector)
+    {
+        $this->injectItems([$collector]);
```
samdark: `injectItems()` expects `Route[]|Group[]` but we're passing `RouteCollectorInterface[]`. Either we need to update phpdoc or adjust code...

src/RouteCollection.php
```diff
@@ -0,0 +1,154 @@
...
+     */
+    public function getRouteTree(bool $routeAsString = true): array
+    {
+        return $this->buildTree($this->items, $routeAsString);
+    }
+
+    /**
+     * Build routes array
+     *
+     * @param Route|Group[] $items
```
samdark:
```suggestion
     * @param Route[]|Group[] $items
```

src/RouteCollectionInterface.php
```diff
@@ -0,0 +1,26 @@
...
+     * @return Route
+     */
+    public function getRoute(string $name): Route;
+
+    /**
+     * Returns routes tree array
+     *
+     * @return array
+     */
+    public function getRouteTree(): array;
```
samdark: It's not a good idea to have `array` as return type in an interface in case certain structure of an array is assumed.

tests/Middleware/RouterTest.php
```diff
@@ -62,6 +62,11 @@ public function __construct(MiddlewareInterface $middleware)
                 $this->middleware = $middleware;
             }
 
+            public function getCurrentRoute(): ?Route
+            {
+                // TODO: Implement getCurrentRoute() method.
```
samdark: ???

tests/RouterFactoryTest.php
```diff
@@ -118,6 +118,11 @@ public function getUriPrefix(): string
                     public function setUriPrefix(string $name): void
                     {
                     }
+
+                    public function getCurrentRoute(): ?Route
+                    {
+                        // TODO: Implement getCurrentRoute() method.
```
samdark: ???

PR #35, opened 2020-02-19
Add container late binding

yiiliveext:
| Q                 | A
| ------------- | ---
| Is bugfix?    | ❌
| New feature?  | ✔️
| Breaks BC?    | ❌
| Tests pass?   | ✔️

Added container late binding/injecting. Just replace `RouteCollectorInterface::class => new GroupFactory()` with `RouteCollectorInterface::class => new Group()` in the your container config and the container will be injected into Route immediately before processing.

### DIFF
--- a/src/MatchingResult.php
+++ b/src/MatchingResult.php
@@ -2,6 +2,7 @@
 
 namespace Yiisoft\Router;
 
+use Psr\Container\ContainerInterface;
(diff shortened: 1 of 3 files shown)


xepozz:
Too magical behavior

yiiliveext:
> Too magical behavior

Why? It is very clear.

samdark:
👍

PR #39, opened 2020-03-27
Fix nested group middlewares injecting bug

yiiliveext:
| Q             | A
| ------------- | ---
| Is bugfix?    | ✔️
| New feature?  | ❌
| Breaks BC?    | ❌
| Tests pass?   | ✔️


samdark:
Note to myself: add the following test.

> Create group A with the route. Create group B. Add group A to group B. Add middleware to group B. Execute route. Ensure middleware is executed.

PR #48, opened 2020-05-09
Update RouteCollectionInterface.php

xepozz:
| Q             | A
| ------------- | ---
| Is bugfix?    | ✔️

PR #49, opened 2020-05-09
Update RouteCollectionInterface.php

xepozz:
| Q             | A
| ------------- | ---
| Is bugfix?    | ✔️

In addition to #48

### DIFF
--- a/src/RouteCollectionInterface.php
+++ b/src/RouteCollectionInterface.php
@@ -20,7 +20,7 @@ interface RouteCollectionInterface
     /**
      * Returns routes tree array
      *
-     * @return array
+     * @return string[]|string[][]|string[][][]|string[][][][]|string[][][][][]
      */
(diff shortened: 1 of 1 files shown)


src/RouteCollectionInterface.php
```diff
@@ -20,7 +20,7 @@ public function getRoute(string $name): Route;
     /**
      * Returns routes tree array
      *
-     * @return array
+     * @return string[]|string[][]|string[][][]|string[][][][]|string[][][][][]
```
samdark, line 23: That looks super-weird. Is there a need for such construct? What does it fix?

xepozz: I would remove it from here and create some `RoutesTreeBuilder`

samdark: OK. Closing the pull request then.

PR #53, opened 2020-06-27
Rename method addMiddleware to withMiddleware

xepozz:
| Q             | A
| ------------- | ---
| Is bugfix?    | ✔️
| New feature?  | ❌
| Breaks BC?    | ✔️
| Tests pass?   | ✔️

I've renamed method because `with` is sign of cloning object


samdark:
No, that is not correct since `with` is like a `set`, it replaces a single value. In this case middleware is actually added i.e. calling method multiple time will add multiple middleware.

Issue #57, opened 2020-07-20
Move executing action out of the router


samdark:
That was proposed by @roxblnfk before in the dev chat. Not sure I've got it exactly but I think this variant is interesting.

## What we have now

1. Router executes handlers: https://github.com/yiisoft/router/blob/master/src/Middleware/Router.php#L43.
2. How it's done can not be customized: https://github.com/yiisoft/router/blob/master/src/Route.php#L373. We have injector there injecting services and that's it.
3. We can't use Yii-specific functionality there since router is not Yii-specific package.
4. There is a potential attribute collision issue between request and infrastructural things such as CSRF middleware. Both are using request attributes. https://github.com/yiisoft/router/blob/master/src/Middleware/Router.php#L40

## Proposal

1. Split getting a matching a route and executing it.
2. Leave getting a route matched in this package, matched route will be set to an attribute of the request.
3. Move route execution into yii-web.

It fixes the following issues:

1. Handler execution is customizable.
2. We can use any Yii-specific things when executing a handler. Here's a good example of what could be done way simpler with such approach: https://github.com/romkatsu/yii-demo/pull/4
3. Attribute collision is easier to avoid (but that's separate thing).

samdark:
Examples of `2` are:

- A custom Request for handler that has extra functionality.
- Ability to return `string` or `[array]` from a handler.

samdark:
Agreed that @roxblnfk would work on proof on concept.

armpogart:
> Ability to return string or [array] from a handler.
Such things are already possible to implement directly in this package and this can be helpful in many cases, see my [PR](https://github.com/yiisoft/router/pull/58).

If I understand it correctly you plan to get rid of handling route directly in this package. In that case I see no any practical usage of this package as a separate package for the community as it doesn't add any considerable value instead of using FastRoute directly for example. I haven't found any php router that is not allowing execution/handling in it. I'm not sure that `Yiisoft\Router` name will be appropriate for this package after the change as it will become just `RouteMatcher`.

What about considering ability to add custom handler to the routes (or group) specified instead of removing route execution/handling functionality altogether. I'm sure adding an ability to define custom handler for routes is very viable and good alternative. It will also add more value to the package and default functionality will remain in here. Any custom handler specified (if specified) will just replace default execution/handling process and the package will just define some interface for custom handlers.

Or maybe I don't fully understand the problem?

roxblnfk:
[...]

PR #58, opened 2020-07-27
Allow returning string from matched routes.

armpogart:
| Q             | A
| ------------- | ---
| Is bugfix?    | ❌
| New feature?  | ✔️
| Breaks BC?    | ❌
| Fixed issues  | comma-separated list of tickets # fixed by the PR, if any

Allow to return string directly from route handlers or callback.

### DIFF
--- a/src/Route.php
+++ b/src/Route.php
@@ -3,6 +3,7 @@
 namespace Yiisoft\Router;
 
 use InvalidArgumentException;
+use Psr\Http\Message\ResponseFactoryInterface;
 use Yiisoft\Http\Method;
 use Yiisoft\Injector\Injector;
 use Psr\Container\ContainerInterface;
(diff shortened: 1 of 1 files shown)


armpogart:
@romkatsu I will add use cases and tests later, it was just a quick PR for now. Let it remain here, I will change the status to draft until progress on #57 after if we still need it, I will rebase and improve PR.

xepozz:
@armpogart you should use yiisoft/data-response to allow the action to be able to return string.

armpogart:
@xepozz Yes, sure when in context of Yii framework or app with dependency on `yiisoft/data-response` package, but I think it may be still helpful to have such simple functionality out of the box in the `yiisoft/router` package to make it more appealing as a standalone package. Just look how simple can example in the readme become with it.

Or maybe the router can depend on `yiisoft/data-response` package (as it is fairly lightweight) and provide all the functionality by default through it.

I'm not strongly insisting to adding such functionality to the package, but I've opened the PR as proof of concept for the comments in #57 and will keep this PR here for discussion before router rewrite.

roxblnfk:
> I'm not strongly insisting to adding such functionality to the package, but I've opened the PR as proof of concept for the comments in #57 and will keep this PR here for discussion before router rewrite.

A few months ago, we had ActionCaller, overriding which, you could change the behavior of the action launch. Now the logic is hardcoded and I think we need to add flexibility and customization, not add hardcode.

samdark:
Won't be merged as is. Closing.

Issue #56, opened 2020-07-10
Remove UrlMatcherInterface from main container


roxblnfk:
I think we should remove [`UrlMatcherInterface`](https://github.com/yiisoft/router/blob/4a8966569805558d58e3a8758a77258237b44fe0/src/UrlMatcherInterface.php) from main container because this is stateful service.

As solutuion we can pass `UrlMatcherInterface` object in a controller with `RequestInterface` as an item of [attributes](https://github.com/php-fig/http-message/blob/master/src/ServerRequestInterface.php#L199-L210). Or as subitem with route params ( #55 )

samdark:
1. Where should URL matcher instance be created?
2. How would you change these?

- https://github.com/yiisoft/app/blob/master/config/params.php#L78
- https://github.com/yiisoft/router-fastroute/blob/master/src/UrlGenerator.php#L28
- https://github.com/yiisoft/router/blob/master/src/Middleware/Router.php#L19
- https://github.com/yiisoft/yii-debug/blob/master/src/Collector/RouterCollector.php#L21
- https://github.com/yiisoft/yii-demo/blob/master/src/Command/Router/ListCommand.php#L21

roxblnfk:
> 1. Where should URL matcher instance be created?

If the Matcher is independent and stateless, then something like this is expected:

```php
interface UrlMatcherInterface
{
    public function match(RouteCollection $routes, ServerRequestInterface $request): MatchingResult;
}
```

But we have Matcher that dependeds on the route. This means that you need to receive it directly from a **router**.
Here, by the **router**, i mean a container that contains all services necessary for routing related to only this router.
Let it be `RouterInterface`, which we don't have.

---

### What should `UrlMatcherInterface` do?

From [description](https://github.com/yiisoft/router/blob/4a8966569805558d58e3a8758a77258237b44fe0/src/UrlMatcherInterface.php#L7-L10):
```php
/**
 * UrlMatcherInterface allows finding a matching route given a PSR-8 server request. It is preferred to type-hint
 * against it in case you need to match URL.
 */
```

1. Why does Matcher include the `getCurrentRoute()` method? Shouldn't it be in the `MatchingResult`?
2. Regarding getLastMatchedRequest method same question.
3. Why do we take the collection of routes from the matcher and not from the router?

---

> 2. How would you change these?
> 
> 1 https://github.com/yiisoft/app/blob/master/config/params.php#L78
[...]

Issue #37, opened 2020-03-19
Light routes


xepozz:
# Problems

 I see errors in the implementation of `Route`:
- The `Route` contains a container.
- `Route` process the `Request` themselves, i.e. in addition to providing information, they also process it.

# What the problems do they make?
- It would seem that having a collection of `Route`'s, it is possible to "dump" (`var_dump` or other), but this will not work with the current implementation, because the `Container` will be dumped, in which there are links to the same `Route` and everything will be repeated in recursion.
- There is no single point of processing Routes. Each `Route` is independent, but treats itself equally.

# Solution

I suggest:
- Return RouterInterface and make an independent Router
- Remove container from Routes
- Remove Request processing from Route
- Process all Routes in the Router
- All middlewars will be initialized in Router too

Current workflow:
```php
$route = new Route('GET', 'home/index');
$response = $route->handle($request);
```

# Working with routes:

Future workflow:
```php
$route1 = new Route('GET', 'home/index');
$route2 = new Route('POST', 'home/index');

$router = new Router($routes); 
// or
$router = $container->get(RouterInterface);
$router->injectRoutes($routes);
// or other ways to create the Router

$response = $router->handle($request);
```

---

# Conclusion

- Now `Route` can be "dumped", exported or cached very easly.
- `Route` "knows" no more than what it needs to know. Now it's just a `DTO`
- A single point of processing all requests (the `Router`). It gives:
   - The controlled process of matching any routes
   - Now you can implement the storage of the current route being processed
   - Other minor

samdark:
Decided not to do it for now. May solve debugging issue with `__debugInfo()`.

samdark:
TODO:

- [ ] Implement `__debugInfo()`.

PR #60, opened 2020-08-16
Router refactoring

armpogart:
| Fixed issues  | #57, #56, #37 

Implementation of the new router. This is still a work in progress. There are many things that are missing, but you can already see the direction of the refactoring and a couple of working test cases.

Some context behind the decisions and incentives of the changes (see also linked issues while I will be updating this description):
1. Almost all classes are implementations of their respective interfaces. This is done to ensure maximum future flexibility in changing any part of the router, be it in standalone package usage or in the context of Yii 3 framework. Maybe some interfaces are redundant, it's not clear yet as the work is not finished.
2. Traits are used extensively to avoid code duplication. There is an IDE problem (with correct autocompletion) which is solved by adding PHPDoc for now. That problem is fully resolved in PHP 8 as `static` return type is added (https://wiki.php.net/rfc/static_return_type). This will fix declaring correct interface method return types for IDEs.
3. `ContainerInterface` dependency is removed from everywhere except `DefaultDispatcher` and `DispatcherInterface` doesn't even force you to have that dependency, so any custom `DispatcherInterface` can decide on it's own what dependencies it needs.
4. All route processing and handling logic is removed from everywhere and is encapsulated in `DispatcherInterface`. Now `DispatcherInterface` implementation will decide how to handle request in case of route match.
5. Almost all mentions of middlewares (e.g. `addMiddleware`) were replace with `handler` as it became technically incorrect to name it middleware and accept anything. Handler type and format support is now dictated by `DispatcherInterface` used as all the processing is done there. As a consequence we can support any new format/type just by replacing `DispatcherInterface` implementation.
6. `Route` in old implementation is divided into 2 separate classes (both must be immutable): `Yiisoft\Router\Route\Definition` and `Yiisoft\Router\Route\Route`. `Definition` contains pure definition (encapsulates that part) of the real world route (path, host, schemes, port, methods supported, accepts supported, route name). And `Route` adds additional logic on top of that, such as handlers and dispatcher. The separation is done to make easier replacing just one part (that part will be `Route` usually). `Definition` won't be used directly (though it's possible to do so). Framework will define own `Route` class to add factories, such as `Route::redirect(...)` and `Route::view(...)`.
    1. I don't know whether we `DefinitionInterface` at all. This is one of the interfaces that may be redundant.
    2. I'm not sure whether the definition must support multiple schemes or just one?
    3. I'm not sure whether name really belongs to `Definition` or `Route`?
7. `RouteCollector` and `RouteCollection` interfaces from the old implementation are combined to one `RouteCollection` interface. It is still WIP.
8. `UrlMatcherInterface` is renamed to `MatcherInterface` and it now doesn't depend on `RouteCollectionInterface`. It has an additional method now `matchForRoutes(iterable $routes, ServerRequestInterface $request)` and will be working with flattened routes directly instead of `RouteCollectionInterface`. I don't remember why I have kept `match(ServerRequestInterface $request)` method and added another one. I guess it will be clear while I do the adapter. I had something in my mind, but am not sure what :)
9. New class `Router` is added (don't confuse with `Router` middleware in old implementation, the middleware will still be there as a separate class, I just need to re-add that) which is combining all the functionality of the router (adding routes, matching, generation, handling). It is especially handy when using package standalone. Yii 3 framework can still use router services separately in its' implementation.
10. `DispatcherInterface` is the part which is responsible for all the handling/processing logic of handlers (middlewares in old implementation) of the matched route. `DispatcherInterface` implementation is used in standalone `Router`, and `Router` middleware for processing/handling under the hood.

Some examples of test cases that are there to showcase the new router:
1. In this case `DefaultDispatcher` is used, which is given to `Router` constructor. `DefaultDispatcher` wraps all callable middlewares to `MiddlewareInterface` and runs them in correct order. It includes all the functionality that was present for processing `Route` before refactoring.
```php
$request = new ServerRequest('GET', '/');
$router = $this->createRouter()
    ->addRoute(Route::get('/', static function () {
        return new Response(200, [], 'test');
    }));
$response = $router->handle($request);

$this->assertSame(200, $response->getStatusCode());
```
2. This one applies custom `DispatcherInterface` to the route used and consequently if the route matches its' custom dispatcher is used. Custom dispatcher just returns 404 response without taking into account any middlewares attached.
```php
$request = new ServerRequest('GET', '/');
$route = Route::get('/', static function () {
    return new Response(200, [], 'test');
})->withDispatcher($this->getCustomDispatcher());
$router = $this->createRouter()->addRoute($route);
$response = $router->handle($request);

$this->assertSame(404, $response->getStatusCode());
```

P.S. I will be update PR description while progressing through it.
P.P.S. I haven't worked on fastroute adapter yet, but as far as I see there won't be any problem with it.

Some planned TODOs (not necessarily done in this PR, can be potentially done separately):
- Immutable and Mutable types are used in mixed manner. I just had no time to cleanup that part. And before that we need to decide which classes will be fully immutable?
- Support serialization of definition, routes and collections through yiisoft/serializer.
- Support route compilation (maybe through serializaiton). The support would be added to respective interfaces and adapter, so that if the adapter supports that feature, it would also have `CompiledRouteMatcher` and `CompiledRouteGenerator` to work with compilation output.
- Remove support of handler format `[SomeHandler::class, 'action']` and replace it with `'SomeHandler::action'` or `'SomeHandler@action'` from `DefaultDispatcher`. Or maybe even just add another format support and keep the old one there.


armpogart:
TODO: add `NotFoundHandler` for the case of usage of the package standalone when no route matches.

samdark:
@armpogart what's left in this pull request?

samdark:
Some ideas from the pull request were implemented. @armpogart do want to continue it?

Issue #66, opened 2020-09-15
Dispatcher's method withMiddlewares has potential bug


denyadzi:
Continuing with my interest in middleware dispatching in Yii3. Here I've noticed a kind of logical problem with Dispatcher::withMiddlewares method. One may expect, that giving a new set of middlewares to it will change the workflow the dispatch-method will do. But it doesn't work in this way due to the handler-stack being built for first set of middlewares only.

### What steps will reproduce the problem?

I've written a little test that demonstrates that https://github.com/denyadzi/router/blob/dispatcher-with-middlewares-issue/tests/DispatcherTest.php

### Possible solution 

Stack should be nulled together with middlewares assignment, but this will make stack being rebuilt all the time, as this method is called in MatchingResult. Possibly [this approach](https://github.com/yiisoft/yii-web/pull/307/commits/e51214bfccc15b6581966c34d2ae9d284f15186a) could be useful here

yiiliveext:
Dispatcher can't be called twice with different middleware sets. Though, the code really looks a little bit strange. I'll look, how we can do it clearer.

yiiliveext:
> 
> ### What steps will reproduce the problem?
> I've written a little test that demonstrates that https://github.com/denyadzi/router/blob/dispatcher-with-middlewares-issue/tests/DispatcherTest.php

This test is incorrect, `Dispatcher` is immutable, so you can't reproduce this case in a real application with a container.
The correct test is:
```php
public function testWithMiddlewares_shouldAffectFlow()
    {
        $middleware200 = $this->createMock(MiddlewareInterface::class);
        $middleware200
            ->method('process')
            ->willReturn(new Response(200));
        $middleware400 = $this->createMock(MiddlewareInterface::class);
        $middleware400
            ->method('process')
            ->willReturn(new Response(400));

        $request = new ServerRequest('GET', '/');
        $fallbackHandler = $this->createMock(RequestHandlerInterface::class);

        //dispatcher that stored in the container;
        $dispatcher = $this->container->get(Dispatcher::class);

        //dispatcher in the first request
        $firstDispatcher = $dispatcher->withMiddlewares([$middleware200]);
        $response200 = $dispatcher->dispatch($request, $fallbackHandler);
        //dispatcher in the second request
        $secondDispatcher = $firstDispatcher->withMiddlewares([$middleware400]);
        $response400 = $secondDispatcher->dispatch($request, $fallbackHandler);

        $this->assertEquals(200, $response200->getStatusCode());
[...]

PR #67, opened 2020-09-16
Fix clonning bug

yiiliveext:
| Q             | A
| ------------- | ---
| Is bugfix?    | ✔️
| New feature?  | ❌
| Breaks BC?    | ❌
| Fixed issues  | #66 

Disadvantage: the stack will be rebuilt every request when we'are using a request loop.

### DIFF
--- a/src/Dispatcher.php
+++ b/src/Dispatcher.php
@@ -47,8 +47,11 @@ final class Dispatcher implements DispatcherInterface
 
     public function withMiddlewares(array $middlewares): DispatcherInterface
     {
+        $stack = $this->stack;
+        $this->stack = null;
         $clone = clone $this;
         $clone->middlewares = $middlewares;
+        $this->stack = $stack;
 
         return $clone;
     }


rustamwin (approved):

src/Dispatcher.php
```diff
@@ -47,8 +47,11 @@ public function dispatch(ServerRequestInterface $request, RequestHandlerInterfac
 
     public function withMiddlewares(array $middlewares): DispatcherInterface
     {
+        $stack = $this->stack;
+        $this->stack = null;
         $clone = clone $this;
         $clone->middlewares = $middlewares;
+        $this->stack = $stack;
```
xepozz, line 54: Isn't it equivalent?
```suggestion
        $clone = clone $this;
        $clone->middlewares = $middlewares;
        $clone->stack = null;
```

yiiliveext: > Isn't it equivalent?

I this case you should change modificator of `stack` property to `protect`.

yiiliveext: Though.... I'm wrong, the private modifier works too.

yiiliveext: Done. Thanks.

yiiliveext:
See https://github.com/yiisoft/router/pull/68
This pr will be closed

PR #68, opened 2020-09-16
Fix dispatcher injecting

yiiliveext:
| Q             | A
| ------------- | ---
| Is bugfix?    | ❌
| New feature?  | ❌
| Breaks BC?    | ✔️


src/Group.php
```diff
@@ -59,18 +59,15 @@ public static function create(?string $prefix = null, $routes = [], DispatcherIn
         return new self($prefix, $callback, $dispatcher);
     }
 
-    public function withDispatcher(DispatcherInterface $dispatcher): self
+    public function injectDispatcher(DispatcherInterface $dispatcher): void
```
samdark, line 62: Can it be immutable?

src/Route.php
```diff
@@ -23,24 +23,26 @@ final class Route
...
+    private array $middlewareDefinitions = [];
     private array $defaults = [];
 
     private function __construct(?DispatcherInterface $dispatcher = null)
     {
         $this->dispatcher = $dispatcher;
     }
 
-    public function withDispatcher(DispatcherInterface $dispatcher): self
+    public function injectDispatcher(DispatcherInterface $dispatcher): void
```
samdark, line 34: Can it be immutable?

yiiliveext: > Can it be immutable?

No, it doesn't. We should reuse the stack on the next request in the request loop.

src/Group.php
```diff
@@ -59,11 +59,11 @@ public static function create(?string $prefix = null, $routes = [], DispatcherIn
         return new self($prefix, $callback, $dispatcher);
     }
 
-    public function withDispatcher(DispatcherInterface $dispatcher): self
+    public function withDispatcher(MiddlewareDispatcher $dispatcher): self
     {
         $group = clone $this;
         $group->dispatcher = $dispatcher;
-        foreach ($group->items as $index => $item) {
+        foreach ($this->items as $index => $item) {
```
samdark, line 66: The change looks suspicious.

samdark:
👍

PR #71, opened 2020-09-29
Add auto response for OPTIONS method

yiiliveext:
| Q             | A
| ------------- | ---
| Is bugfix?    | ❌
| New feature?  | ✔️
| Breaks BC?    | ❌

### DIFF
--- a/src/Middleware/Router.php
+++ b/src/Middleware/Router.php
@@ -9,6 +9,7 @@ use Psr\Http\Message\ResponseInterface;
 use Psr\Http\Message\ServerRequestInterface;
 use Psr\Http\Server\MiddlewareInterface;
 use Psr\Http\Server\RequestHandlerInterface;
+use Yiisoft\Http\Method;
 use Yiisoft\Http\Status;
 use Yiisoft\Router\MiddlewareDispatcher;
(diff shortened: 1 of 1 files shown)


src/Middleware/Router.php
```diff
@@ -18,6 +19,7 @@ final class Router implements MiddlewareInterface
     private UrlMatcherInterface $matcher;
     private ResponseFactoryInterface $responseFactory;
     private MiddlewareDispatcher $dispatcher;
+    private bool $optionsAutoReponse = false;
```
rustamwin, line 22:
```suggestion
    private bool $optionsAutoResponse = false;
```

samdark: Should it be optional?

samdark:
That needs to be default and work well with both custom OPTIONS handler/middleware.

src/Middleware/Router.php
```diff
@@ -26,11 +28,24 @@ public function __construct(UrlMatcherInterface $matcher, ResponseFactoryInterfa
         $this->dispatcher = $dispatcher;
     }
 
+    public function withOptionsAutoResponse(): self
```
rustamwin, line 31: What about `withPreflight()`?

PR #74, opened 2020-10-17
Fix config

xepozz:
| Q             | A
| ------------- | ---
| Is bugfix?    | ✔️
| New feature?  | ❌
| Breaks BC?    | ❌
| Fixed issues  |

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -46,7 +46,7 @@
             "dev-master": "3.0.x-dev"
         },
         "config-plugin": {
-            "common": "config/common.php"
+            "web": "config/web.php"
         }
     },
     "config": {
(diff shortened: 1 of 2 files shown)


composer.json
```diff
@@ -46,7 +46,7 @@
             "dev-master": "3.0.x-dev"
         },
         "config-plugin": {
-            "common": "config/common.php"
```
samdark, line 49: Wouldn't that break https://github.com/yiisoft/yii-demo/blob/master/src/Command/Router/ListCommand.php?

samdark:
Can't be applied anymore since config changed.

PR #91, opened 2020-12-28
Fallback router concept

samdark:
| Q             | A
| ------------- | ---
| Is bugfix?    | ❌
| New feature?  | ✔️
| Breaks BC?    | ❌
| Fixed issues  | https://github.com/yiisoft/app/issues/46

### DIFF
--- a/src/FallbackGenerator.php
+++ b/src/FallbackGenerator.php
@@ -1,0 +1,47 @@
+<?php
+
+declare(strict_types=1);
+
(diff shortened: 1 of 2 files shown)


samdark:
Likely that should be a separate framework-specific package because it will require reflection to determine if handler is available.

samdark:
Also it is likely that the approach itself will coflict with grouping classes by use-case since it requires a certain determined classes layout.

PR #92, opened 2020-12-28
Add route override

yiiliveext:
| Q             | A
| ------------- | ---
| Is bugfix?    | ❌
| New feature?  | ✔️
| Breaks BC?    | ✔️


samdark (approved):

roxblnfk:
I would expect an exception to be thrown when defining routes of the same name.
Maybe we need to make it customizable?

samdark:
👍

Issue #100, opened 2021-03-27
Error with multiple routes files: prependMiddleware() can not be used before action()


Mister-42:
### What steps will reproduce the problem?
```php
return [
    RouteCollectionInterface::class => static function (ApplicationParameters $appParams, RouteCollectorInterface $collector) use ($config) {
        $collector
            ->middleware(CsrfMiddleware::class)
            ->middleware(FormatDataResponse::class)
            ->addGroup(
                Group::create()->routes(...$config->get("routes-{$appParams->site}")),
            )
            ->addGroup(
                Group::create()->routes(...$config->get("routes-{$appParams->siteReversed}"))
            );

        return new RouteCollection($collector);
    },
];
```

I am not using `prependMiddleware` in my config and/or config anywhere. For testing purposes I have commented out every `middleware` in my routes as well.

### What do you get instead?
`prependMiddleware() can not be used before action().`

### Additional info

| Q                | A
| ---------------- | ---
| Version          | dev
| PHP version      | 8.0.3
| Operating system | Debian 10

yiiliveext:
Any route must contains action() method.

Mister-42:
> Any route must contains action() method.

Why?
```php
        Route::get('orig')
            ->middleware(fn (Redirect $redirect) => $redirect->toRoute('new')),
```

What would be the action here?

yiiliveext:
> > Any route must contains action() method.
> 
> Why?
> 
> ```
>         Route::get('orig')
>             ->middleware(fn (Redirect $redirect) => $redirect->toRoute('new')),
> ```
> 
> What would be the action here?

Redirect is the action.

Mister-42:
> Redirect is the action.

No, that is middleware and this rule failes with above very abstract error unless an empty action is added.

Mister-42:
Also, this should be completely valid for its purpose:

```php
    // Assets for urlGenerator, requests will not be handled by Yii
    Route::get('/images/{file}')
        ->host($appParams->assetDomain)
        ->name('img'),
```

yiiliveext:
> Also, this should be completely valid for its purpose:
> 
> ```
>     // Assets for urlGenerator, requests will not be handled by Yii
>     Route::get('/images/{file}')
>         ->host($appParams->assetDomain)
>         ->name('img'),
> ```

I don't think this is the right way to use routes.

Mister-42:
@samdark Would that be a valid use case?

yiiliveext:
[...]

PR #101, opened 2021-03-29
Add static routes

yiiliveext:
| Q             | A
| ------------- | ---
| Is bugfix?    | ❌
| New feature?  | ✔️
| Breaks BC?    | ❌
| Fixed issues  | #100

### DIFF
--- a/src/Route.php
+++ b/src/Route.php
@@ -33,11 +33,17 @@ final class Route
 
     public function injectDispatcher(MiddlewareDispatcher $dispatcher): void
     {
+        if ($this->isStatic()) {
+            throw new RuntimeException('Static route can not has middleware dispatcher.');
+        }
         $this->dispatcher = $dispatcher;
     }
 
     public function withDispatcher(MiddlewareDispatcher $dispatcher): self
     {
+        if ($this->isStatic()) {
+            throw new RuntimeException('Static route can not has middleware dispatcher.');
+        }
         $route = clone $this;
         $route->dispatcher = $dispatcher;
         return $route;
@@ -142,6 +148,19 @@ final class Route
 
     /**
      * @param string $pattern
+     *
+     * @return self
+     */
+    public static function static(string $pattern): self
+    {
+        $route = new self();
+        $route->pattern = $pattern;
+        $route->isStatic = true;
+        return $route;
+    }
+
+    /**
+     * @param string $pattern
      * @param MiddlewareDispatcher|null $dispatcher
      *
      * @return self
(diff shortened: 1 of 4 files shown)


Mister-42:
Wouldn't Route::static make more sense?

yiiliveext:
> Wouldn't Route::static make more sense?

What do you mean? 
The example from your issue
```php
//Assets for urlGenerator, requests will not be handled by Yii
    Route::get('/images/{file}')
        ->static() 
        ->host($appParams->assetDomain)
        ->name('img'),
```
That's all. You can use static routes in a common routes file/array.

Mister-42:
I meant 

```php
    //Assets for urlGenerator, requests will not be handled by Yii
    Route::static('/images/{file}')
        ->host($appParams->assetDomain)
        ->name('img'),
```

Sorry, I commented without enough information.

yiiliveext:
> I meant
> 
> ```
>     //Assets for urlGenerator, requests will not be handled by Yii
>     Route::static('/images/{file}')
>         ->host($appParams->assetDomain)
>         ->name('img'),
> ```
> 
> Sorry, I commented without enough information.

Maybe it makes sense. Need to review from more devs.

samdark:
There was an ability to make URL rule with matching only or generation only in Yii 2: https://github.com/yiisoft/yii2/blob/master/framework/web/UrlRule.php#L38

I haven't personally used it myself so I'm not sure about use cases.

yiiliveext:
> I meant
> 
> ```
>     //Assets for urlGenerator, requests will not be handled by Yii
>     Route::static('/images/{file}')
>         ->host($appParams->assetDomain)
>         ->name('img'),
> ```

Done

tests/RouteTest.php
```diff
@@ -124,6 +124,47 @@ public function testDefaults(): void
...
+        Route::static('/')->middleware(fn () => 1);
+    }
+
+    public function testStaticFailWihPrependMiddleware(): void
+    {
+        $this->expectExceptionMessage("Static route can not has middleware.");
+        Route::static('/')->middleware(fn () => 1);
+    }
+
+    public function testStaticFailWihAddAction(): void
```
samdark, line 162:
```suggestion
    public function testStaticFailWithAddAction(): void
```

src/Route.php
```diff
@@ -139,6 +146,19 @@ public static function options(string $pattern, ?MiddlewareDispatcher $dispatche
         return self::methods([Method::OPTIONS], $pattern, $dispatcher);
     }
 
+    /**
+     * @param string $pattern
+     *
+     * @return self
+     */
+    public static function static(string $pattern): self
```
samdark, line 154: There is no need for a separate method. Generation-only route is the one that doesn't have action.

samdark:
👍

Issue #102, opened 2021-04-04
Use parameter bag for Route::get*


samdark:
## Problem

We have a group of getters in Route: https://github.com/yiisoft/router/blob/master/src/Route.php#L305. These are used internally but are never used by end user polluting IDE autocomplete list.

## Proposal

Introduce parameter bag and a method such as `getParameters()`.

PR #108, opened 2021-06-20
Use parameter bag

rustamwin:
| Q             | A
| ------------- | ---
| Is bugfix?    | ❌
| New feature?  | ❌
| Breaks BC?    | ✔️
| Fixed issues  | #102

### DIFF
--- a/src/Route.php
+++ b/src/Route.php
@@ -13,21 +13,39 @@ use Yiisoft\Middleware\Dispatcher\MiddlewareDispatcher;
  */
 final class Route
 {
-    private ?string $name = null;
+    public const PARAMETER_NAME = 'name';
+    public const PARAMETER_METHODS = 'methods';
+    public const PARAMETER_PATTERN = 'pattern';
+    public const PARAMETER_HOST = 'host';
+    public const PARAMETER_DEFAULTS = 'defaults';
+    public const PARAMETER_OVERRIDE = 'override';
+    public const PARAMETER_ACTION_ADDED = 'action_added';
+
     /** @var string[] */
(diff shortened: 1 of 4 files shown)


src/Route.php
```diff
@@ -13,21 +13,39 @@
...
 final class Route
 {
-    private ?string $name = null;
+    public const PARAMETER_NAME = 'name';
+    public const PARAMETER_METHODS = 'methods';
+    public const PARAMETER_PATTERN = 'pattern';
+    public const PARAMETER_HOST = 'host';
+    public const PARAMETER_DEFAULTS = 'defaults';
+    public const PARAMETER_OVERRIDE = 'override';
+    public const PARAMETER_ACTION_ADDED = 'action_added';
```
samdark, line 22:
```suggestion
    public const NAME = 'name';
    public const METHODS = 'methods';
    public const PATTERN = 'pattern';
    public const HOST = 'host';
    public const DEFAULTS = 'defaults';
    public const OVERRIDE = 'override';
```

src/Route.php
```diff
@@ -13,21 +13,39 @@
...
+
     /** @var string[] */
-    private array $methods;
-    private string $pattern;
-    private ?string $host = null;
-    private bool $override = false;
+    private array $parameters;
     private ?MiddlewareDispatcher $dispatcher;
-    private bool $actionAdded = false;
```
samdark, line 23: It should stay private boolean flag since it was never used externally.

samdark (approved):

samdark:
Thanks.

Issue #109, opened 2021-06-24
Two interfaces instead of parameter bag


samdark:
There's a better idea (more performant, event more perfect API):

1. Revert https://github.com/yiisoft/router/pull/108
2. Introduce two interfaces  RouterInterface (everything except getters), RouteParametersInterface (getters).
3. Use RouterInterface for end-user stuff. Use RouteParametersInterface internally.

PR #110, opened 2021-06-25
Revert parameters bag and introduce interfaces

rustamwin:
| Q             | A
| ------------- | ---
| Is bugfix?    | ❌
| New feature?  | ❌
| Breaks BC?    | ✔️
| Fixed issues  | #109


samdark (approved):

roxblnfk (approved):
But i think it will be good to restore `getParameters(): RouteParametersInterface` method in the `RouteInterface`

PR #113, opened 2021-07-03
Improve Group, add Group::name, Group::host

rustamwin:
| Q             | A
| ------------- | ---
| Is bugfix?    | ❌
| New feature?  | ✔️
| Breaks BC?    | ?
| Fixed issues  | -


samdark (approved):

samdark:
@rustamwin would you please merge master into the PR branch so tests pass?

src/RouteCollection.php
```diff
@@ -127,6 +127,12 @@ private function injectGroup(Group $group, array &$tree, string $prefix = ''): v
 
             /** @var Route $modifiedItem */
             $modifiedItem = $item->pattern($prefix . $item->getPattern());
+            if ($group->getHost() !== null && $modifiedItem->getHost() === null) {
+                $modifiedItem = $modifiedItem->host($group->getHost());
+            }
+            if ($group->getName() !== null && strpos($modifiedItem->getName(), implode(', ', $modifiedItem->getMethods())) === false) {
```
vjik, line 133: What is the mean of checking `strpos($modifiedItem->getName(), implode(', ', $modifiedItem->getMethods())) === false` ?

rustamwin: Checking that the route has a default name. We skip unnamed routes.

vjik: May be create method `$route->isDefaultName(): bool` ?

rustamwin: > May be create method $route->isDefaultName(): bool ?

What do you guys think @yiisoft/yii3?

src/GroupParametersInterface.php
```diff
@@ -0,0 +1,21 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Yiisoft\Router;
+
+interface GroupParametersInterface
```
vjik, line 7: Why? Has use case for read groups in application?

rustamwin: Deleted.

src/RouteCollection.php
samdark: Seems to be alright.

samdark (approved):

PR #118, opened 2021-08-18
Improve CurrentRoute, minor fixes

rustamwin:
| Q             | A
| ------------- | ---
| Is bugfix?    | ✔️/❌
| New feature?  | ✔️/❌
| Breaks BC?    | ✔️/❌
| Fixed issues  | -

See https://github.com/yiisoft/router-fastroute/pull/79

### DIFF
--- a/src/CurrentRoute.php
+++ b/src/CurrentRoute.php
@@ -11,6 +11,11 @@ use Psr\Http\Message\UriInterface;
  */
 final class CurrentRoute
 {
+    /**
+     * Current Route
+     *
+     * @var RouteParametersInterface|null
+     */
     private ?RouteParametersInterface $route = null;
     /**
      * Current URI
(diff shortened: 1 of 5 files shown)


src/CurrentRoute.php
```diff
@@ -39,16 +44,23 @@ public function getUri(): ?UriInterface
...
 
+    /**
+     * @param RouteParametersInterface $route
+     */
     public function setRoute(RouteParametersInterface $route): void
     {
-        $this->route = $route;
+        if ($this->route === null) {
+            $this->route = $route;
+        }
```
yiiliveext, line 54: Otherwise, throw an exception

src/CurrentRoute.php
```diff
@@ -39,16 +44,23 @@ public function getUri(): ?UriInterface
...
-     * @param UriInterface|null $uri
+     * @param UriInterface $uri
      */
-    public function setUri(?UriInterface $uri): void
+    public function setUri(UriInterface $uri): void
     {
-        $this->uri = $uri;
+        if ($this->uri === null) {
+            $this->uri = $uri;
+        }
```
yiiliveext, line 64: Otherwise, throw an exception

yiiliveext:
You should add a resetter for `CurrentRoute` in the container config. Also, IMHO, a good idea is to add CurrentRoute to the request attribute.

samdark (approved):

PR #119, opened 2021-08-19
Add `CurrentRouteInterface` containing only getters for use in userland

vjik:
| Q             | A
| ------------- | ---
| Is bugfix?    | ❌
| New feature?  | ✔️
| Breaks BC?    | ❌
| Fixed issues  | -

Adopt PRs:
- https://github.com/yiisoft/router-fastroute/pull/78
- https://github.com/yiisoft/app/pull/180
- https://github.com/yiisoft/demo/pull/333


yiiliveext:
I prefer this way https://github.com/yiisoft/router/pull/118

vjik:
> I prefer this way #118

This PR not does not contradict to #118

yiiliveext:
> > I prefer this way #118
> 
> This PR not does not contradict to #118

It is not necessary if we accept #118

src/CurrentRouteInterface.php
```diff
@@ -0,0 +1,24 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Yiisoft\Router;
+
+use Psr\Http\Message\UriInterface;
+
+interface CurrentRouteInterface
```
xepozz, line 9: It's not a `route` at all. Can you please rename it to `provider`?

vjik: Does this mean that `CurrentRoute` also should be renamed to `CurrentRouteProvider` ?

xepozz: It would be great

samdark (approved):

PR #121, opened 2021-09-26
Add getName method

rustamwin:
| Q             | A
| ------------- | ---
| Is bugfix?    | ❌
| New feature?  | ✔️
| Breaks BC?    | ❌
| Fixed issues  | -

### DIFF
--- a/src/CurrentRoute.php
+++ b/src/CurrentRoute.php
@@ -26,6 +26,16 @@ final class CurrentRoute
     private ?UriInterface $uri = null;
 
     /**
+     * Returns the current Route name
+     *
+     * @return string
+     */
+    public function getName(): string
+    {
+        return $this->route === null ? '' : $this->route->getName();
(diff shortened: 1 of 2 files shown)


src/CurrentRoute.php
```diff
@@ -25,6 +25,16 @@ final class CurrentRoute
...
     private ?UriInterface $uri = null;
 
+    /**
+     * Returns the current Route name
+     *
+     * @return string
+     */
+    public function getName(): string
+    {
+        return $this->route === null ? '' : $this->route->getName();
```
vjik, line 35: Maybe better return `?string`?

My logic:
_empty string_ — route exist, but empty.
`null` — route not exist.

rustamwin: Agree. The code was moved from demo layout :)

vjik (approved):

PR #122, opened 2021-09-27
Always add head method to get routes

rustamwin:
| Q             | A
| ------------- | ---
| Is bugfix?    | ✔️/❌
| New feature?  | ✔️/❌
| Breaks BC?    | ✔️/❌
| Fixed issues  | See https://github.com/nikic/FastRoute#a-note-on-head-requests

### DIFF
--- a/src/Route.php
+++ b/src/Route.php
@@ -73,7 +73,7 @@ final class Route implements RouteInterface, RouteParametersInterface
      */
     public static function get(string $pattern, ?MiddlewareDispatcher $dispatcher = null): RouteInterface
     {
-        return self::methods([Method::GET], $pattern, $dispatcher);
+        return self::methods([Method::GET, Method::HEAD], $pattern, $dispatcher);
     }
 
     /**


src/Route.php
```diff
@@ -73,7 +73,7 @@ public function hasDispatcher(): bool
      */
     public static function get(string $pattern, ?MiddlewareDispatcher $dispatcher = null): RouteInterface
     {
-        return self::methods([Method::GET], $pattern, $dispatcher);
+        return self::methods([Method::GET, Method::HEAD], $pattern, $dispatcher);
```
samdark, line 76: From https://github.com/nikic/FastRoute#a-note-on-head-requests:

> Finally, note that applications MAY always specify their own HEAD method route for a given resource to bypass this behavior entirely.

That's an interesting point for the following code:

```php
Route::get('/bla', 'getBlaHandler');
Route::head('/bla', 'getBlaHeadHandler')
```

With the change in this PR we won't be able to do it, I think. But that might be a non-issue...

samdark: Likely should be done in fastroute adapter instead.

rustamwin:
See https://github.com/yiisoft/router-fastroute/pull/82

Issue #55, opened 2020-07-10
Group route parameters in a Request attributes


roxblnfk:
I believe that it would be more correct if the router writes route parameters to its group in the attributes, and not as it is now

```php
Route::get('/test/{id:\w+}/{param1}[/{param2}]', ...);
```

Now, if we request the attributes of the request, we get
```php
[
    // Route params
    'id' => ...,
    'param1' => ...,
    'param2' => ...,
    // From middlewares
    'language' => 'en',
    'format' => 'json',
    '...' => ...
]
```
What happens if we set a competing `format` parameter in the route?

When implementing my proposal:
```php
[
    // Route params
    'router' => [
        'id' => ...,
        'param1' => ...,
        'param2' => ...,
    ]
    // From middlewares
    'language' => 'en',
    'format' => 'json',
    '....' => ...
]
```
No conflicts. And we can always separate route parameters from others

samdark:
I like and dislike the idea. It resolves potential conflicts (and there's no doubt there are going to be conflicts since both sides are developer-powered) but it makes working with request parameters less appealing:

```php
public function actionConfirm(ServerRequestInterface $request)
{
    $id = $request->getAttribute('router', [])['id'] ?? null;
    $code = $request->getAttribute('router', [])['code'] ?? null;
    // ...
}
```

vs current:

```php
public function actionConfirm(ServerRequestInterface $request)
{
    $id = $request->getAttribute('id');
    $code = $request->getAttribute('code');
    // ...
}
```

viktorprogger:
@roxblnfk Can you explain please the usecase for this?

samdark:
We may add ability to get matched arguments from CurrrentRoute. Then we can remove writing arguments to request.

PR #128, opened 2021-11-13
Move parameters group into current route

rustamwin:
| Q             | A
| ------------- | ---
| Is bugfix?    | ❌
| New feature?  | ❌
| Breaks BC?    | ✔️
| Fixed issues  | #55


src/CurrentRouteInterface.php
```diff
@@ -40,7 +40,7 @@ public function getParameters(): array;
      * Returns the current route parameter.
      *
      * @param string $name
-     * @param null $default
+     * @param mixed|null $default
```
vjik, line 43:
```suggestion
     * @param mixed $default
```

src/CurrentRoute.php
```diff
@@ -77,4 +82,23 @@ public function setUri(UriInterface $uri): void
...
+        return $this->parameters[$name] ?? $default;
+    }
+
+    public function setParameters(array $parameters)
+    {
+        if ($this->parameters === []) {
+            $this->parameters = $parameters;
+            return;
+        }
+        throw new RuntimeException('Can not set parameters since it was already set.');
```
vjik, line 102: May be throw `LogicException`?

From docs:

`LogicException` that represents error in the program logic. This kind of exception should lead directly to a fix in your code.

`RuntimeException` thrown if an error which can only be found on runtime occurs.

tests/CurrentRouteTest.php
```diff
@@ -37,4 +37,36 @@ public function testGetCurrentUri(): void
...
+        $this->assertSame('bar', $currentRoute->getParameter('foo'));
+    }
+
+    public function testGetParameterWithNonExist()
+    {
+        $currentRoute = new CurrentRoute();
+        $currentRoute->setParameters(['test' => 1]);
+
+        $this->assertNull($currentRoute->getParameter('foo'));
+    }
```
vjik, line 71: Also need test `getParameter()` with custom default value.

tests/CurrentRouteTest.php
```diff
@@ -37,4 +38,74 @@ public function testGetCurrentUri(): void
...
+
+    public function testSetRouteTwice()
```
devanych, line 82:
```php
public function testSetRouteTwice(): void
```

tests/CurrentRouteTest.php
```diff
@@ -37,4 +38,74 @@ public function testGetCurrentUri(): void
...
+    {
+        $this->expectException(LogicException::class);
+        $this->expectExceptionMessage('Can not set route since it was already set.');
+
+        $currentRoute = new CurrentRoute();
+        $currentRoute->setRoute(Route::get('')->name('test'));
+        $currentRoute->setRoute(Route::get('/home')->name('home'));
+    }
+
+    public function testSetUriTwice()
```
devanych, line 92:
```php
public function testSetUriTwice(): void
```

tests/CurrentRouteTest.php
```diff
@@ -37,4 +38,74 @@ public function testGetCurrentUri(): void
...
+    {
+        $this->expectException(LogicException::class);
+        $this->expectExceptionMessage('Can not set URI since it was already set.');
+
+        $currentRoute = new CurrentRoute();
+        $currentRoute->setUri(new Uri(''));
+        $currentRoute->setUri(new Uri('home'));
+    }
+
+    public function testSetParametersTwice()
```
devanych, line 102:
```php
public function testSetParametersTwice(): void
```

src/CurrentRouteInterface.php
```diff
@@ -28,4 +28,21 @@ public function getRoute(): ?RouteParametersInterface;
      * @return UriInterface|null The current URI.
      */
     public function getUri(): ?UriInterface;
+
+    /**
+     * Returns the current route parameters.
+     *
+     * @return array
+     */
+    public function getParameters(): array;
```
roxblnfk, line 37: If we have a route with path `/{foo}/{bar}` and it matched with `/hello/world` request URI then:
 - parematers are `['foo', 'bar']`
 - arguments are `['foo' => 'hello', 'bar' => 'world']`

I hope the code is consistent with the terminology

src/CurrentRoute.php
```diff
@@ -24,6 +24,11 @@ final class CurrentRoute implements CurrentRouteInterface
      */
     private ?UriInterface $uri = null;
 
+    /**
+     * Current Route parameters
```
samdark, line 28:
```suggestion
     * Current Route parameters.
```

samdark (approved):

src/CurrentRouteInterface.php
samdark: PSR-7 uses "params" so, I think, we should be consistent with it.

roxblnfk: It is not the same case. User has already configured the parameters in the route like custom function.
PSR-7 has nothing to do with it at all.

src/CurrentRoute.php
```diff
@@ -24,6 +24,11 @@ final class CurrentRoute implements CurrentRouteInterface
      */
     private ?UriInterface $uri = null;
 
+    /**
+     * Current Route parameters.
+     */
+    private array $parameters = [];
```
samdark, line 30:
```suggestion
    /**
     * Current Route arguments.
     */
    private array $arguments = [];
```

src/CurrentRoute.php
```diff
@@ -75,6 +80,25 @@ public function setUri(UriInterface $uri): void
...
             return;
         }
-        throw new RuntimeException('Can not set URI since it was already set.');
+        throw new LogicException('Can not set URI since it was already set.');
+    }
+
+    public function getParameters(): array
+    {
+        return $this->parameters;
+    }
```
samdark, line 89:
```suggestion
    public function getArguments(): array
    {
        return $this->arguments;
    }
```

src/CurrentRoute.php
```diff
@@ -75,6 +80,25 @@ public function setUri(UriInterface $uri): void
...
+
+    public function getParameter(string $name, $default = null): ?string
+    {
+        return $this->parameters[$name] ?? $default;
+    }
```
samdark, line 94:
```suggestion
    public function getArgument(string $name, $default = null): ?string
    {
        return $this->arguments[$name] ?? $default;
    }
```

src/CurrentRoute.php
```diff
@@ -75,6 +80,25 @@ public function setUri(UriInterface $uri): void
...
+
+    public function setParameters(array $parameters): void
+    {
+        if ($this->parameters === []) {
+            $this->parameters = $parameters;
+            return;
+        }
+        throw new LogicException('Can not set parameters since it was already set.');
     }
```
samdark, line 103:
```suggestion
    public function setArguments(array $arguments): void
    {
        if ($this->arguments === []) {
            $this->arguments = $arguments;
            return;
        }
        throw new LogicException('Can not set arguments since they were already set.');
    }
```

samdark (commented):
I think @roxblnfk is right about parameters vs arguments:

> The term parameter (sometimes called formal parameter) is often used to refer to the variable as found in the function definition, while argument (sometimes called actual parameter) refers to the actual input supplied at function call.

i.e.

> Parameter is variable in the declaration of function.
> Argument is the actual value of this variable that gets passed to function.

README.md
```diff
@@ -196,6 +198,23 @@ final class MyClass {
 }
 ```
 
+Current route arguments could be obtained the following:
```
samdark, line 201:
```suggestion
Current route arguments could be obtained as follows:
```

PR #135, opened 2021-12-07
Add pre-flight flag to routes

rustamwin:
| Q             | A
| ------------- | ---
| Is bugfix?    | ❌
| New feature?  | ✔️
| Breaks BC?    | ❌

### DIFF
--- a/src/Route.php
+++ b/src/Route.php
@@ -258,6 +258,18 @@ final class Route implements RouteInterface, RouteParametersInterface
         return $route;
     }
 
+    /**
+     * @return self
+     */
+    public function preFlight(): RouteInterface
+    {
+        $route = clone $this;
+        if (!in_array(Method::OPTIONS, $this->methods, true)) {
+            $route->methods[] = Method::OPTIONS;
+        }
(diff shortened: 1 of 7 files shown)


src/Route.php
```diff
@@ -258,6 +258,18 @@ public function disableMiddleware($middlewareDefinition): RouteInterface
...
+    /**
+     * @return self
+     */
+    public function preFlight(): RouteInterface
+    {
+        $route = clone $this;
+        if (!in_array(Method::OPTIONS, $this->methods, true)) {
+            $route->methods[] = Method::OPTIONS;
+        }
+        return $route;
```
vjik, line 270:
```suggestion
        if (!in_array(Method::OPTIONS, $this->methods, true)) {
            $route = clone $this;
            $route->methods[] = Method::OPTIONS;
            return $route;
        }
        return $this;
```

yiiliveext:
This is a candidate for the Darwin Award.

samdark:
https://github.com/yiisoft/router/issues/136

PR #137, opened 2021-12-15
Optimize auto options

samdark:
| Q             | A
| ------------- | ---
| Is bugfix?    | ✔️
| New feature?  | ❌
| Breaks BC?    | ❌
| Fixed issues  | https://github.com/yiisoft/router/pull/137

### DIFF
--- a/src/RouteCollection.php
+++ b/src/RouteCollection.php
@@ -113,8 +113,8 @@ final class RouteCollection implements RouteCollectionInterface
             }
 
             if ($item instanceof Group) {
-                if ($group->hasAutoOptions()) {
-                    $item = $item->withAutoOptions(...$group->getAutoOptions());
+                if ($group->hasCorsMiddleware()) {
+                    $item = $item->withCors($group->getCorsMiddleware());
                 }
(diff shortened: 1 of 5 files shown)


src/RouteCollection.php
```diff
@@ -151,32 +151,33 @@ private function injectGroup(Group $group, array &$tree, string $prefix = '', st
...
+                && !(($pattern === $modifiedItem->getPattern() && $host === $modifiedItem->getHost())
+                    || in_array(Method::OPTIONS, $modifiedItem->getMethods(), true))
+            ) {
+                $pattern = $modifiedItem->getPattern();
+                $host = $modifiedItem->getHost();
+                /** @var Route $optionsRoute */
+                $optionsRoute = Route::options($pattern);
+                if ($host !== null) {
+                    $optionsRoute = $optionsRoute->host($host);
+                }
```
samdark, line 166: Could be moved out of `foreach`.

samdark: `$optionsRoute` could be prepared and then cloned.

Issue #131, opened 2021-11-24
Psalm error on use `RouteCollectorInterface::addGroup()`


vjik:
`RouteCollectorInterface::addGroup()` accepts parameter with type `Group`, but `Group::create()` returns `GroupInterface`:

![image](https://user-images.githubusercontent.com/525501/143257286-214ba2f2-5da7-4b53-97d4-c36965e9dd75.png)

samdark:
Should use interface if possible.

vjik:
No, we don't can use interface in `addGroup()` because need object with getters, but interface contain only setters.

PR #138, opened 2021-12-22
Fix #130 #131: Remove interfaces & getters

### LINKED ISSUE #130: Add a warning about the prohibition of use group and route interfaces
`GroupInterface`, `RouteInterface` and `RouteParametersInterface` has one implementation each in package and shouldn't implement somewhere else. This interfaces need only for convient IDE autocomplete.

I think this warning should be in phpdoc of interfaces.

rustamwin:
| Q             | A
| ------------- | ---
| Is bugfix?    | ❌
| New feature?  | ✔️/❌
| Breaks BC?    | ✔️
| Fixed issues  | #130, #131


src/Group.php
```diff
@@ -12,8 +12,13 @@
 use function in_array;
 use function is_object;
 
-final class Group implements GroupInterface
+final class Group
 {
+    public const PREFIX = 'prefix';
```
samdark, line 17: Not sure these constants are needed because they are in the middle of autocomplete list when defining routes.

src/Group.php
```diff
@@ -127,56 +143,76 @@ public function middleware($middlewareDefinition): GroupInterface
...
+    public function getData(string $key)
     {
-        return $this->items;
-    }
-
-    public function getPrefix(): ?string
-    {
-        return $this->prefix;
-    }
+        $dataMap = $this->getDataMap();
```
samdark, line 199: Good old `switch` would be more suitable here.

src/Route.php
```diff
@@ -11,8 +11,15 @@
 /**
  * Route defines a mapping from URL to callback / name and vice versa.
  */
-final class Route implements RouteInterface, RouteParametersInterface
+final class Route
 {
+    public const NAME = 'name';
```
samdark, line 16: Not sure these are needed. We don't want people to use these.

src/Route.php
```diff
@@ -277,34 +310,31 @@ public function __toString(): string
...
+    /**
+     * @param string $key
+     *
+     * @return mixed
+     */
+    public function getData(string $key)
     {
-        return $this->host;
-    }
+        $dataMap = $this->getDataMap();
```
samdark, line 320: `switch` would be more suitable.

src/Route.php
```diff
@@ -277,34 +310,31 @@ public function __toString(): string
...
 
-    public function isOverride(): bool
-    {
-        return $this->override;
+        return $dataMap[$key] ?? null;
```
samdark, line 322: It is better to throw an exception instead of returning null in this case.

src/RouteCollection.php
```diff
@@ -82,9 +82,9 @@ private function injectItem($route): void
             return;
         }
 
-        $this->items[] = $route->getName();
-        $routeName = $route->getName();
-        if (isset($this->routes[$routeName]) && !$route->isOverride()) {
+        $this->items[] = $route->getData('name');
+        $routeName = $route->getData('name');
```
yiiliveext, line 86:
```suggestion
        $routeName = $route->getData('name'); 
        $this->items[] = $routeName;        
```

src/CurrentRoute.php
```diff
@@ -60,9 +110,11 @@ public function getUri(): ?UriInterface
     }
 
     /**
-     * @param RouteParametersInterface $route
+     * @param Route $route
+     *
+     * @internal
      */
-    public function setRoute(RouteParametersInterface $route): void
+    public function setRoute(Route $route): void
```
vjik, line 117: Suggestion: merge methods `setRoute()` and `setArguments` to one method, because in router they used in one place one after the other.

src/Group.php
```diff
@@ -127,58 +138,82 @@ public function middleware($middlewareDefinition): GroupInterface
...
+                return $this->corsMiddleware;
+            case 'items':
+                return $this->items;
+            default:
+                throw new InvalidArgumentException('Unknown data key: ' . $key);
+        }
     }
 
+    /**
+     * @return array
```
vjik, line 213: Suggestion: get middleware definitions also via `getData()`

src/Route.php
```diff
@@ -277,34 +309,31 @@ public function __toString(): string
...
-
-    public function getDefaults(): array
+    /**
+     * @param string $key
+     *
+     * @return mixed
+     *
+     * @internal
+     */
+    public function getData(string $key)
```
vjik, line 319: Suggestion: check has middlewares also via `getData()`

.phpstorm.meta.php/Group.php
```diff
@@ -0,0 +1,17 @@
+<?php
+
+namespace PHPSTORM_META {
+
+    use Yiisoft\Router\Group;
```
vjik, line 5: PhpStorm don't support section `use`.

.phpstorm.meta.php/Route.php
```diff
@@ -0,0 +1,18 @@
+<?php
+
+namespace PHPSTORM_META {
+
+    use Yiisoft\Router\Route;
```
vjik, line 5: PhpStorm don't support section `use`.

vjik (changes requested):

samdark (approved):

Issue #139, opened 2021-12-22
Increase psalm level


vjik:
Minimum up to 2 level. Ideally up to 1

PR #140, opened 2021-12-24
Psalm level 1 + minor refacroting and improvements

vjik:
| Q             | A
| ------------- | ---
| Is bugfix?    | ❌
| New feature?  | ❌
| Breaks BC?    | ❌
| Fixed issues  | #139

### DIFF
--- a/src/MatchingResult.php
+++ b/src/MatchingResult.php
@@ -8,19 +8,20 @@ use Psr\Http\Message\ResponseInterface;
 use Psr\Http\Message\ServerRequestInterface;
 use Psr\Http\Server\MiddlewareInterface;
 use Psr\Http\Server\RequestHandlerInterface;
+use RuntimeException;
 use Yiisoft\Http\Method;
 use Yiisoft\Middleware\Dispatcher\MiddlewareDispatcher;
 
 final class MatchingResult implements MiddlewareInterface
 {
-    private bool $success;
-    private Route $route;
+    private ?Route $route;
     private array $arguments = [];
     private array $methods = [];
     private ?MiddlewareDispatcher $dispatcher = null;
 
-    private function __construct()
+    private function __construct(?Route $route)
     {
+        $this->route = $route;
     }
 
     public function withDispatcher(MiddlewareDispatcher $dispatcher): self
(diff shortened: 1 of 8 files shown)


src/Route.php
```diff
@@ -335,10 +364,15 @@ public function __debugInfo()
 
     private function getDispatcherWithMiddlewares(): MiddlewareDispatcher
     {
+        if ($this->dispatcher === null) {
+            throw new RuntimeException('Route do not container middleware dispatcher.');
```
rustamwin, line 368:
```suggestion
            throw new RuntimeException('There is no dispatcher on the route.');
```

src/MatchingResult.php
```diff
@@ -30,45 +40,61 @@ public function withDispatcher(MiddlewareDispatcher $dispatcher): self
...
+     */
     public function methods(): array
     {
         return $this->methods;
     }
 
     public function route(): Route
     {
+        if ($this->route === null) {
+            throw new RuntimeException('Route not available on matching failure.');
```
rustamwin, line 95:
```suggestion
            throw new RuntimeException('There is no route to match.');
```

src/MatchingResult.php
```diff
@@ -30,45 +40,61 @@ public function withDispatcher(MiddlewareDispatcher $dispatcher): self
...
+     */
     public function methods(): array
     {
         return $this->methods;
     }
 
     public function route(): Route
     {
+        if ($this->route === null) {
+            throw new RuntimeException('There is no route to match.');
```
samdark, line 95:
```suggestion
            throw new RuntimeException('There is no route in the matching result.');
```

src/Route.php
```diff
@@ -335,10 +364,15 @@ public function __debugInfo()
 
     private function getDispatcherWithMiddlewares(): MiddlewareDispatcher
     {
+        if ($this->dispatcher === null) {
+            throw new RuntimeException('There is no dispatcher on the route.');
```
samdark, line 368:
```suggestion
            throw new RuntimeException(sprintf('There is no dispatcher in the route %s', $this->getData('name')));
```

PR #156, opened 2021-12-28
New locales based on pattern

### LINKED ISSUE #158: Add query parameters to `UrlGeneratorInterface`
```php
public function generate(
  string $name,
  array $arguments = [],
  array $queryParameters = []
 ): string;
 
public function generateAbsolute(
  string $name,
  array $arguments = [],
  array $queryParameters = [],
  string $scheme = null,
  string $host = null
): string;

public function generateFromCurrent(
  array $replacedArguments,
  ?string $fallbackRouteName = null,
  array $queryParameters = []
): string;
```

yiiliveext:
| Q             | A
| ------------- | ---
| Is bugfix?    | ❌
| New feature?  | ❌
| Breaks BC?    | ❌

Fix #158

### DIFF
--- a/src/UrlGeneratorInterface.php
+++ b/src/UrlGeneratorInterface.php
@@ -45,6 +45,8 @@ interface UrlGeneratorInterface
         string $host = null
     ): string;
 
+    public function generateCurrent(array $replacedParams): string;
+
     public function getUriPrefix(): string;
 
     public function setUriPrefix(string $name): void;
--- a/src/UrlGeneratorInterface.php
+++ b/src/UrlGeneratorInterface.php
@@ -4,6 +4,8 @@ declare(strict_types=1);
 
 namespace Yiisoft\Router;
 
+use Stringable;
+
 /**
  * UrlGeneratorInterface allows generating URL given route name and parameters.
  * It is preferred to type-hint against it in case you need to generate a URL.
@@ -16,11 +18,11 @@ interface UrlGeneratorInterface
      * @param string $name Name of the route.
      * @param array $parameters Parameter-value set.
      *
-     * @throws RouteNotFoundException In case there is no route with the name specified.
-     *
      * @return string URL generated.
      *
-     * @psalm-param array<string,null|object|scalar> $parameters
+     * @psalm-param array<string,null|Stringable|scalar> $parameters
+     *@throws RouteNotFoundException In case there is no route with the name specified.
+     *
      */
     public function generate(string $name, array $parameters = []): string;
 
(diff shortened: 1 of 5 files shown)


src/UrlGeneratorInterface.php
```diff
@@ -45,13 +45,11 @@ public function generateAbsolute(
         string $host = null
     ): string;
 
+    public function generateCurrent(array $replacedParams, string $fallbackRouteName = null): string;
```
samdark, line 48: How about adjusting name a bit?

```suggestion
    public function generateFromCurrent(array $replacedParams, string $fallbackRouteName = null): string;
```

src/UrlGeneratorInterface.php
```diff
@@ -45,13 +45,11 @@ public function generateAbsolute(
...
     public function getUriPrefix(): string;
 
     public function setUriPrefix(string $name): void;
 
-    public function getLocales(): array;
-
-    public function setLocales(array $locales): void;
-
-    public function setLocaleParameterName(string $localeParameterName): void;
+    public function setDefault(string $name, $value): void;
```
rustamwin, line 54: Is there case for non-string value?
```suggestion
    public function setDefault(string $name, string $value): void;
```

samdark: That's done for convenience not to cast to string every time.

src/UrlGeneratorInterface.php
```diff
@@ -45,13 +45,11 @@ public function generateAbsolute(
         string $host = null
     ): string;
 
+    public function generateFromCurrent(array $replacedParams, string $fallbackRouteName = null): string;
```
vjik, line 48:
```suggestion
    public function generateFromCurrent(array $replacedParameters, ?string $fallbackRouteName = null): string;
```

src/UrlGeneratorInterface.php
```diff
@@ -45,13 +45,11 @@ public function generateAbsolute(
...
     public function getUriPrefix(): string;
 
     public function setUriPrefix(string $name): void;
 
-    public function getLocales(): array;
-
-    public function setLocales(array $locales): void;
-
-    public function setLocaleParameterName(string $localeParameterName): void;
+    public function setDefault(string $name, $value): void;
```
vjik, line 54:
```suggestion
    /**
     * @param null|object|scalar $value
     */
    public function setDefault(string $name, $value): void;
```

vjik, line 54: May be allow setmultiple parameters as `setDefaults(array $defaults)` ?
This is will be consistent: in other places of this packages we set arrays of parameters/defaults.

src/UrlGeneratorInterface.php
yiiliveext: Done

src/UrlGeneratorInterface.php
yiiliveext: This can be not a string. This method is not only for locales.

src/UrlGeneratorInterface.php
yiiliveext: No, there are many places where we can set different defaults. The one method overrides previous defaults.

samdark (approved):

PR #157, opened 2021-12-28
Fix #142: Consider ignored middlewares in hasMiddlewares

### LINKED ISSUE #142: `Route::getData('hasMiddlewares')` ignore disabled middlewares
For example:

```php
        $route = Route::get('/')
            ->middleware(DummyMiddleware1::class)
            ->disableMiddleware(DummyMiddleware1::class);

        $this->assertTrue($route->getData('hasMiddlewares'));
```
        
Seems, in this case `$route->getData('hasMiddlewares')` should returns `false`.

dehbka:
| Q             | A
| ------------- | ---
| Is bugfix?    | ✔️
| New feature?  | ❌
| Breaks BC?    |❌
| Fixed issues  | https://github.com/yiisoft/router/issues/142

### DIFF
--- a/src/Route.php
+++ b/src/Route.php
@@ -334,7 +334,7 @@ final class Route
             case 'dispatcherWithMiddlewares':
                 return $this->getDispatcherWithMiddlewares();
             case 'hasMiddlewares':
-                return $this->middlewareDefinitions !== [];
+                return $this->hasMiddlewares();
             case 'hasDispatcher':
                 return $this->dispatcher !== null;
             default:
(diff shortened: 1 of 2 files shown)


yiiliveext:
It doesn't make sense. A non-static route always has at least one middleware (action). A static route doesn't have any middleware, so there is nothing to disable.

dehbka:
So issue doesn't make sense?)))

samdark:
Kind of:

1. It is now internal thing.
2. The route still has middleware despite it not being executed.
3. I was too fast in judging original issue :(

So, thanks for contribution but won't merge it and will close the issue.

Issue #159, opened 2022-04-12
Add multi-host support


rustamwin:
Route can currently set one host

samdark:
What's the use-case for multiple hosts?

rustamwin:
Sometimes you need support multiple hosts, but FastRoute doesn't support pattern like this:
`{_host:(yiiframework.com|yf.com)}`

samdark:
OK.

PR #163, opened 2022-06-07
Add multi hosts

Gerych1984:
| Q             | A
| ------------- | ---
| Is bugfix?    | ❌
| New feature?  | ✔️
| Breaks BC?    | ❌
| Fixed issues  | #159

1. Allow several hosts in `Route::host` method
2. Add method `Route::isMultiHost`


src/Route.php
```diff
@@ -186,13 +186,27 @@ public function pattern(string $pattern): self
     /**
      * @return self
      */
-    public function host(string $host): self
+    public function host(string ...$hosts): self
```
samdark, line 189: That looks a bit weird in routes:

```php
->host('demo.test', 'demo.com')
```

According to English grammar it should be hosts, not host. Maybe we need an alias method?

```suggestion
    public function hosts(string ...$hosts): self
    {
        // call host() here
    }

    public function host(string ...$hosts): self
```

src/Route.php
```diff
@@ -369,7 +390,8 @@ public function __debugInfo()
             'name' => $this->name,
             'methods' => $this->methods,
             'pattern' => $this->pattern,
-            'host' => $this->host,
+            'host' => $this->getData('host'),
+            'hosts' => $this->hosts,
```
samdark, line 394: Having both host and hosts looks unnecessary.

samdark (approved):

samdark:
Looks good. Need a line for CHANGELOG.

CHANGELOG.md
```diff
@@ -2,8 +2,8 @@
 
 ## 1.0.1 under development
 
-- no changes in this release.
+- Allow multiple separate hosts with new `Route::hosts` method.
```
rustamwin, line 5:
```suggestion
- Enh #163: Allow multiple separate hosts with new `Route::hosts` method.
```

samdark:
Thanks!

PR #203, opened 2023-05-15
Add `RouteArgument` attribute for Yii Hydrator

vjik:
| Q             | A
| ------------- | ---
| Is bugfix?    | ❌
| New feature?  | ✔️
| Breaks BC?    | ❌
| Fixed issues  | -


samdark (approved):

xepozz (changes requested):
`Route` is a wrong name for this kind of attribute.

`Route` is already reserved for routing: the class and coming attribute as well. 
I suggest you think about `RouteAttribute` or something similar. Do not confuse developers with one more `Route` entity.

xepozz:
Why is it `argument`, isn't `parameter` better?

vjik:
> Why is it `argument`, isn't `parameter` better?

In router, this is called "arguments": https://github.com/yiisoft/router/blob/0892cbebfdfbe73ae9d2a64a133c96daf1552c7a/src/CurrentRoute.php#L112

composer.json
```diff
@@ -39,6 +39,7 @@
         "vimeo/psalm": "^4.30|^5.20",
         "yiisoft/di": "^1.0",
         "yiisoft/dummy-provider": "^1.0.0",
+        "yiisoft/hydrator": "dev-master",
```
samdark, line 42:
```suggestion
        "yiisoft/hydrator": "^1.0",
```

samdark (approved):

PR #220, opened 2023-10-09
Add routes resource & attributes

rustamwin:
| Q             | A
| ------------- | ---
| Is bugfix?    | ❌
| New feature?  | ✔️
| Breaks BC?    | ✔️/❌
| Fixed issues  | comma-separated list of tickets # fixed by the PR, if any


src/Resource/ResourceInterface.php
```diff
@@ -0,0 +1,19 @@
...
+
+namespace Yiisoft\Router\Resource;
+
+use Yiisoft\Router\Group;
+use Yiisoft\Router\Route;
+
+/**
+ * `ResourceInterface` is a resource of routes.
+ */
+interface ResourceInterface
```
vjik, line 13:
```suggestion
interface RoutesProviderInterface
```

src/Resource/AttributeResource.php
```diff
@@ -0,0 +1,61 @@
...
+
+namespace Yiisoft\Router\Resource;
+
+use olvlvl\ComposerAttributeCollector\Attributes;
+use Yiisoft\Router\Group;
+use Yiisoft\Router\Route;
+
+/**
+ * An attribute resource represents routes that declared via PHP Attributes.
+ * @codeCoverageIgnore
```
vjik, line 13:
```suggestion
 * @codeCoverageIgnore
```

Why ignore?

src/RouteCollector.php
```diff
@@ -25,6 +32,12 @@ public function addRoute(Route|Group ...$routes): RouteCollectorInterface
         return $this;
     }
 
+    public function addResource(ResourceInterface $resource): RouteCollectorInterface
```
vjik, line 35: It need to add resources in runtime? If no, better pass resources array to constructor.

src/Resource/AttributeResource.php
```diff
@@ -0,0 +1,61 @@
...
+     */
+    private static array $reflectionsCache = [];
+
+    public function getRoutes(): array
+    {
+        $routes = [];
+        $groupRoutes = [];
+        $routePredicate = Attributes::predicateForAttributeInstanceOf(Route::class);
+        $targetMethods = Attributes::filterTargetMethods($routePredicate);
+        foreach ($targetMethods as $targetMethod) {
```
vjik, line 28: Is it parse all PHP files? Can we search in a specific folder?

src/RouteCollector.php
rustamwin: Yes, it will be added in runtime like routes.

src/Resource/AttributeResource.php
rustamwin: Yes, it's a composer plugin, can be configured in composer.json. We may change this implementation in the future (perhaps when `yiisoft/classifier` is ready).

PR #221, opened 2023-10-15
Add `RouteAttributeInterface`

vjik:
| Q             | A
| ------------- | ---
| Is bugfix?    | ❌
| New feature?  | ❌
| Breaks BC?    | ✔️
| Fixed issues  | -

### DIFF
--- a/src/Route.php
+++ b/src/Route.php
@@ -15,8 +15,7 @@ use function in_array;
 /**
  * Route defines a mapping from URL to callback / name and vice versa.
  */
-#[Attribute(Attribute::TARGET_METHOD | Attribute::TARGET_CLASS | Attribute::IS_REPEATABLE)]
-class Route implements Stringable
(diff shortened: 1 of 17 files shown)


src/Route.php
```diff
@@ -15,8 +14,7 @@
 /**
  * Route defines a mapping from URL to callback / name and vice versa.
  */
-#[Attribute(Attribute::TARGET_METHOD | Attribute::TARGET_CLASS | Attribute::IS_REPEATABLE)]
-class Route implements Stringable
+final class Route implements Stringable
```
rustamwin, line 17: Maybe we leave it as attribute like `Group`?

rustamwin (approved):
Overall good 👍

src/Route.php
vjik: This class has parameter "action" in constructor that is not needed in attribute.

Issue #193, opened 2023-03-14
Named arguments vs immutable methods


xepozz:
What's the point to create an object every method calls?

Let's take a look at the `Route` class:
<img width="304" alt="image" src="https://user-images.githubusercontent.com/6815714/224928164-5e9e9824-b2eb-4941-9f2f-067fa774584a.png">

The coolest thing is you can use auto-complete feature to call any of these methods. Is it all?
I think so. Let's imagine an average corporate app with more than 100 routes.
Every application bootstrap it create this 100+ routes. But if we count it comes to 100*3.5 created objects:
100 route * 3-4 immutable methods calls, so 350 created object every bootstrap just for the one matched route in the end. That's awful.
If we open `Route`'s constructor and let users create it instead we can decrease the created objects numbers back to 100.
```php
class Route {
    public function __construct(
        private string $pattern,
        private array $methods,
        private string $name = null,
        private array $middlewares = [],
}
```
That's enough to gain performance easy.

So the way to use `Route` changes as the following:
```php
Route::get('/')
  ->name('site/index')
  ->action([SiteController::index, 'index'])
  ->middlewares([AuthMiddleware::class])
;
```
```php
new Route(
  pattern: '/',
  name: 'site/index',
  action: [SiteController::index, 'index'],
  middlewares: [AuthMiddleware::class],
);
```

All of these properties can be `readonly`, but it's better to later it for the next release. Also all these methods are questionable. Do we really need keep both named arguments and methods?

Due to this changes #179 turns to be real just by one line above the class `#[\Attribute]`

xepozz:
Cheap costs are still greater than zero

samdark:
Need to measure these in order to make a decision about if it makes sense to change routing syntax.

vjik:
I like named arguments. More shortly, but without loss of readability.

xepozz:
I told a few times long ago that `Route` should be just a dto, but now it's kind of god object and not the best one.

Issue #171, opened 2022-07-04
Add friendly exception with examples for exception "middleware() can not be used after action()"


vjik:

samdark:
Obsoleted by 7be2f65

PR #225, opened 2023-11-09
New syntax

rustamwin:
| Q             | A
| ------------- | ---
| Is bugfix?    | ❌
| New feature?  | ✔️
| Breaks BC?    | ✔️
| Fixed issues  | #193, #171

Current syntax

```diff

- use Yiisoft\Router\Route;
- use Yiisoft\Router\Group;
+ use Yiisoft\Router\Builder\RouteBuilder as Route;
+ use Yiisoft\Router\Builder\GroupBuilder as Group;


Group::create()->routes(
    Route::get('')->name('index'),
-   Route::get('/posts')->name('posts'),
+   Route::get('/posts')->name('posts'), // No getters, only setters
);

### CODE REVIEW

**Review by vjik — CHANGES_REQUESTED** [review]
1) #191 suggest use named arguments, and this must improve performance. In this PR keep routes/groups object and additional add new routes/groups objects with named arguments. Seems, this decrease performance...

2) This not fix #171.

**Comment by rustamwin** [comment]
> In this PR keep routes/groups object and additional add new routes/groups objects with named arguments. Seems, this decrease performance...

We can mention it in docs or drop the current syntax.

> This not fix https://github.com/yiisoft/router/issues/171.

Yes, doesn't fix directly. Since there is no exception, there is no need for it.

**Comment by vjik** [comment]
> We can mention it in docs or drop the current syntax.

But what goal of syntax change?

**Review by vjik — CHANGES_REQUESTED** [review]
Can we make `Route` and `Group` constructors public and keep BC? 

I think, `GroupBuilder` and `RouteBuilder` are not needed.

**Comment by rustamwin** [comment]
> Mark as deprecated, then remove in major version.

Let's abandon right away since the PR has major changes.

**Comment by samdark** [comment]
@rustamwin resolving of conflicts is needed.

File: README.md [hunk 1]
```diff
@@ -49,8 +49,8 @@ Common usage of the router looks like the following:
 
 ```php
 use Yiisoft\Router\CurrentRoute;
-use Yiisoft\Router\Group;
-use Yiisoft\Router\Route;
+use Yiisoft\Router\Builder\GroupBuilder as Group;
+use Yiisoft\Router\Builder\RouteBuilder as Route;
```
samdark: We can avoid changing syntax by introducing static facades into old classes.

`Route::get()`, `post()`, `methods()`, etc. would return `RouteBuilder`, while `Group::create()` would return `GroupBuilder`. The public constructors would still create mutable data objects:

```php
$route = new Route([Method::GET], '/users');
```

samdark:
I also think we can keep backwards compatibility and mark some methods as deprecated but not sure it's needed.

PR #295, opened 2026-08-13
Fix #295: Deprecate static route and group factories in favor of constructors, Fix #193: Add public constructors for routes and groups and method-specific route classes

### LINKED ISSUE #193: Named arguments vs immutable methods

samdark:
method-specific route classes

| Q             | A
| ------------- | ---
| Is bugfix?    | ❌
| New feature?  | ✔️
| Breaks BC?    | ❌
| Fixed issues  | #193


src/Route.php
```diff
@@ -47,12 +44,48 @@ final class Route implements Stringable
...
+     * @param array[]|callable[]|string[] $middlewares Handler middleware definitions that should be invoked for a matched route.
+     * @param array<string,null|Stringable|scalar> $defaults Parameter default values indexed by parameter names.
+     * @param string[] $hosts Hosts that the route applies to.
+     * @param bool $override Whether the route should replace an existing route with the same name.
+     * @param array[]|callable[]|string[] $disabledMiddlewares Middleware definitions to exclude when the action is handled.
      */
-    private function __construct(
+    public function __construct(
         private array $methods,
         private string $pattern,
```
vjik, line 62: I suggest move `$pattern` to first place. Seems, it will be more convenient to use.

src/Route.php
vjik:
```suggestion
    private array $middlewares;
```

vjik:
```suggestion
    private array $disabledMiddlewares;
```

src/Route.php
```diff
@@ -47,12 +44,48 @@ final class Route implements Stringable
     private array $defaults = [];
```
vjik:
```suggestion
    private array $defaults;
```

src/Route.php
```diff
@@ -47,12 +44,48 @@ final class Route implements Stringable
...
+        /** @infection-ignore-all Array keys are discarded by MiddlewareFilter::filter(). */
+        $this->disabledMiddlewares = array_values($disabledMiddlewares);
+
+        foreach ($hosts as $host) {
+            $host = rtrim($host, '/');
+
+            if ($host !== '' && !in_array($host, $this->hosts, true)) {
+                $this->hosts[] = $host;
+            }
+        }
```
vjik: Duplicate code of `hosts()` method. Extract to private `setHosts()` method.

src/Route.php
```diff
@@ -47,12 +44,48 @@ final class Route implements Stringable
...
+        array|callable|string|null $action = null,
+        array $middlewares = [],
+        array $defaults = [],
+        array $hosts = [],
+        private bool $override = false,
+        array $disabledMiddlewares = [],
+    ) {
+        /** @infection-ignore-all Array keys are discarded by MiddlewareFilter::filter(). */
+        $this->middlewares = array_values($middlewares);
+        $this->defaults = array_map(strval(...), $defaults);
```
vjik: Duplicate code of `defaults()` method. Extract to private `setDefaults()` method.

src/Group.php
```diff
@@ -39,9 +33,40 @@ final class Group
...
+        $this->disabledMiddlewares = array_values($disabledMiddlewares);
+        $this->corsMiddleware = $corsMiddleware;
+
+        foreach ($hosts as $host) {
+            $host = rtrim($host, '/');
+
+            if ($host !== '' && !in_array($host, $this->hosts, true)) {
+                $this->hosts[] = $host;
+            }
+        }
```
vjik: Duplicate code of `hosts()` method. Extract to private `setHosts()` method.

vjik (changes requested):

src/Group.php
```diff
@@ -212,6 +199,20 @@ public function getData(string $key): mixed
...
+    /**
+     * @param string[] $hosts
+     */
+    private function setHosts(array $hosts): void
+    {
+        foreach ($hosts as $host) {
+            $host = rtrim($host, '/');
+
+            if ($host !== '' && !in_array($host, $this->hosts, true)) {
+                $this->hosts[] = $host;
```
vjik: `setHosts()` should override value. Current implementation always adds items. A test for this case is also needed.

vjik (changes requested):

vjik (approved):

src/Route.php
```diff
@@ -133,7 +162,15 @@ public static function options(string $pattern): self
      */
     public static function methods(array $methods, string $pattern): self
     {
-        return new self($methods, $pattern);
+        if (static::class !== self::class) {
```
samdark, line 195: That's a guard against `Get::post()` usage. We can't avoid the method since it's there in a parent class already.

src/Route.php
```diff
@@ -316,6 +375,42 @@ public function getData(string $key): mixed
         };
     }
 
+    /**
+     * @param string[] $hosts
+     *
+     * @return string[]
+     */
+    public static function normalizeHosts(array $hosts): array
```
vjik: `normalizeHosts` is not needed as public interface. What about extract it to separate internal class?

CHANGELOG.md
```diff
@@ -2,6 +2,9 @@
 
 ## 4.0.3 under development
 
+- Chg: Deprecate static route and group factories in favor of constructors (@samdark)
```
vjik:
```suggestion
- Chg #295: Deprecate static route and group factories in favor of constructors (@samdark)
```

CHANGELOG.md
```diff
@@ -2,6 +2,9 @@
...
+
```
vjik:
```suggestion
```

README.md
```diff
@@ -124,64 +128,84 @@ application middleware processes the request.
...
+    name: 'page-add',
+    action: [PageController::class, 'actionAdd'],
+);
 ```
 
-If you want to generate a URL based on route and its parameters, give it a name with `name()`. Check "Creating URLs"
-for details.
+The static factories `Route::get()`, `post()`, `put()`, `delete()`, `patch()`, `head()`, `options()`,
+`methods()`, and `Group::create()` are deprecated. Use constructors instead. The immutable configuration methods
+remain available; when chaining them after `new`, wrap the construction in parentheses for PHP 8.1 compatibility.
```
vjik:
```suggestion
remain available.
```

I think the constructor parameters should be sufficient, and in most cases there will be no need to call the methods.

README.md
```diff
@@ -124,64 +128,84 @@ application middleware processes the request.
...
 
-`action()` in the above is a primary middleware definition that is invoked last when matching result `process()`
+If you want to generate a URL based on route and its parameters, provide the `name` constructor argument.
+Check "Creating URLs" for details.
```
vjik: Add link

UPGRADE.md
```diff
@@ -3,6 +3,27 @@
 This file contains the upgrade notes for the Yii Router.
 These notes highlight changes that could break your application when you upgrade it from one major version to another.
 
+## 4.0.3
```
vjik: No need. UPGRADE.md describes only backward-incompatible changes.

README.md
```diff
@@ -233,6 +261,21 @@ and `disableMiddleware()`. These middleware are executed prior to matched route'
...
+
+new Group(
+    prefix: '/api',
+    routes: [new Get('/comments')],
+    middlewares: [ApiAuthentication::class],
+    hosts: ['https://example.com'],
+    namePrefix: 'api/',
+);
+```
+
```
vjik:
```suggestion
```

I think this example is not needed. Example above already use named args.

vjik (approved):

